PR #6, opened 2019-05-30
Update firebase rules

pjanik:
Open access to resource JSONs for signed in users.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -9,5 +9,8 @@ service cloud.firestore {
     match /{document=**} {
       allow read, write: if false;
     }
+    match /sources/{source}/resources/{document=**} {
+      allow read: if request.auth.uid != null && request.auth.token.firebase.sign_in_provider != 'anonymous';
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


firestore.rules
```diff
@@ -9,5 +9,8 @@ service cloud.firestore {
     match /{document=**} {
       allow read, write: if false;
     }
+    match /sources/{source}/resources/{document=**} {
+      allow read: if request.auth.uid != null && request.auth.token.firebase.sign_in_provider != 'anonymous';
```
dougmartin: @pjanik are we planning to narrow these back down use the class hash or class id?

pjanik: I don't think we want to do that for "resources" => activity/sequence structure.
Activity structure doesn't contain any private data. We could be concerned about students getting access to it and checking correct answers, but using class_hash won't help with that anyway. 

The plan was to use clash_hash or class_url for rules related to the student answers.

pjanik:
@dougmartin, this never has been approved or merged, so I'm using the same PR for another small update.

pjanik:
@dougmartin, okay, another update, this time a final one. Could you please take a look? Teachers can read all answers of students in their class, students can read only their own answers.

PR #7, opened 2019-06-18
Remove answers from run record.

knowuh:
The Run objects saved in FireStore should not include the answers.

This  PR also adds Jest testing framework.

Extracts  a function for extracting and removing answers from the run. Test that function.

[#166766021]

https://www.pivotaltracker.com/story/show/166766021


knowuh:
@dougmartin  -- Pretty small change actually, just some noise because I added jest testing.

PR #25, opened 2020-07-10
Added learner query upload [#172979222]

dougmartin:
Also updates resource api endpoint to query for url instead of looking up by id.


dougmartin:
@scytacki i've updated the code based on your comments.  Once it is approved I'll update the staging SAM stack and the Firebase  functions.  I'm looking around to try to see how we can tag/version releases to both places.

scytacki:
It was already approved so merge at will.

PR #29, opened 2020-08-04
Increase api gateway timeout from 3 seconds to 300 seconds [#173924040]

dougmartin:
The "internal server error" we saw was actually not caused by the function but rather by the api gateway timing out.  The default timeout is 3 seconds, I increased this to 300 (5 minutes).


dougmartin:
@scytacki this is live and accessible from https://learn-report.staging.concord.org/report/learner.

PR #30, opened 2020-08-07
Update learner and anonymous answer rules

sfentress:
This modifies the answer rules such that

1. A learner can now update and create content, but only within their current class
2. Unauthenticated users can read, update and create any content with a run key. I arbitrary required a minimum of 10 characters for the run-key, to prevent any code that might accidentally use an empty string, or something non-random like the class id.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -22,7 +22,7 @@ service cloud.firestore {
               && string(request.auth.token.platform_user_id) == string(request.resource.data.platform_user_id)
               && request.auth.token.class_hash == request.resource.data.context_id;
       // anonymous requests based on run_key, which must be larger than 10 chars
-      allow read, update: if resource.data.run_key.size() > 10;
+      allow read, update: if resource.data.run_key.size() > 10 && resource.data.run_key == request.resource.data.run_key;
       allow create: if request.resource.data.run_key.size() > 10;
     }  
 
--- a/firestore.rules
+++ b/firestore.rules
@@ -16,10 +16,10 @@ service cloud.firestore {
       allow read: if request.auth.token.user_type == 'teacher' && request.auth.token.class_hash == resource.data.context_id;
       allow read: if request.auth.token.user_type == 'learner' && string(request.auth.token.platform_user_id) == string(resource.data.platform_user_id);
       allow update: if request.auth.token.user_type == 'learner'
-      				&& string(request.auth.token.platform_user_id) == string(resource.data.platform_user_id)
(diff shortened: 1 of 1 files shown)


firestore.rules
```diff
@@ -13,8 +13,17 @@ service cloud.firestore {
       allow read: if request.auth.uid != null && request.auth.token.firebase.sign_in_provider != 'anonymous';
     }
     match /sources/{source}/answers/{document=**} {
```
kswenson, line 15: Having just been in this space for the CLUE teacher support work, let me double-check my understanding here. The `{document=**}` means the rules apply to any document below `answers` at any hierarchical depth. Several of the rules apply to the content of the document, e.g. `string(resource.data.platform_user_id)`. Presumably this means that every document below `answers` has those fields, which I'm guessing is not a particularly deep or granular hierarchy.

firestore.rules
```diff
@@ -13,8 +13,17 @@ service cloud.firestore {
...
+      allow read: if request.auth.token.user_type == 'teacher' && request.auth.token.class_hash == resource.data.context_id;
+      allow read: if request.auth.token.user_type == 'learner' && string(request.auth.token.platform_user_id) == string(resource.data.platform_user_id);
+      allow update: if request.auth.token.user_type == 'learner'
+              && string(request.auth.token.platform_user_id) == string(resource.data.platform_user_id)
+              && request.auth.token.class_hash == resource.data.context_id;
+      allow create: if request.auth.token.user_type == 'learner'
+              && string(request.auth.token.platform_user_id) == string(request.resource.data.platform_user_id)
+              && request.auth.token.class_hash == request.resource.data.context_id;
+      // anonymous requests based on run_key, which must be larger than 10 chars
+      allow read, update: if resource.data.run_key.size() > 10;
```
kswenson, line 25: Don't we also want to match the `run_key` in this case, e.g. something like `request.resource.data.run_key == resource.data.run_key`?

firestore.rules
pjanik: > which I'm guessing is not a particularly deep or granular hierarchy.

Yes, we just have flat answer JSON documents, there are no subcollections at the moment. And yes, we make assumptions about some fields being available, but it seems fine to me.

firestore.rules
pjanik: I'd second that, it seems this would be useful to have this check here, so one anonymous user can't update somebody else's answer.

sfentress: D'oh, of course you're right.

PR #33, opened 2020-09-14
Fixed two errors found when doing prod deployment

dougmartin:
1. The body of the try/catch error wrapper needs to be a string (it was an object which caused the "Internal server error" message)
2. The rethrown error in the getResource Firebase call referenced an unknown variable

### DIFF
--- a/query-creator/create-query/app.js
+++ b/query-creator/create-query/app.js
@@ -69,10 +69,10 @@ exports.lambdaHandler = async (event, context) => {
     console.log(err);
     return {
       statusCode: 500,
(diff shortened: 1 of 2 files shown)


dougmartin:
@scytacki / @knowuh I tagged both of you but just need a review from 1 of you.  I'd like to deploy this early tomorrow morning so that I can show it at the demo meeting.

knowuh:
@dougmartin — looks good to me.

knowuh (approved):
Looks good.

PR #38, opened 2021-01-19
update rules

scytacki:
Update the rules to be more locked down.
This checks that the platform_id is not set when the run_key is used.
And it makes sure the platform_id is required when logged in users are accessing documents.

This also adds tests for all of these updated rules. 

TODO:

- [x] document the changes to the rules better
- [x] add tests for `/plugin_states`
- [x] add tests for `/user_settings`
- [x] add tests for `/question_feedbacks`
- [x] add tests for `/activity_feedbacks`
- [x] add tests for `/feedback_settings`
- [x] add github actions to run the rules tests on each check in

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -36,10 +36,10 @@ service cloud.firestore {
 
     // Note: this conditional is used in read rules, which do not require a
     // matching context_id, so the context_id is not enforced here
-    function learnerOwner() {
+    function learnerOwner(res) {
       return request.auth.token.user_type == 'learner'
-        && string(request.auth.token.platform_user_id) == string(resource.data.platform_user_id)
-        && request.auth.token.platform_id == resource.data.platform_id;
+        && string(request.auth.token.platform_user_id) == string(res.data.platform_user_id)
+        && request.auth.token.platform_id == res.data.platform_id;
     }
 
     function notChangingIdentifyingProps() {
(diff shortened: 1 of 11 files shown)


scytacki:
@dougmartin I added you as a reviewer of this, before it is finished. 
I think the rules are all set. At this point I'm just adding the tests to verify that. 
The tests did turn up at least one thing I missed so there might be more.

I think I'll work on making this run in GitHub Actions, and then I'll try to refactor the tests so they can be applied to both the `/answers` path and the `/plugin_states`.

I don't think there is an easy way to send the coverage information to codecov, so I think the coverage checking will have to be manual.

tests/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+Some useful example tests: https://github.com/zkohi/firebase-testing-samples/blob/sign_in_provider/__tests__/firestore.rules.test.ts
+
+Useful note about the `auth` object passed to initializeTestApp:
+https://github.com/firebase/firebase-tools/issues/2405#issuecomment-651315898
+
+The `auth` object is actually the `token` in the rules.
+
+In order to use the Emulator UI to look at the data in the database from the test runs,
+the projectId in the tests, has to match the project id in .firebaserc.
```
kswenson, line 42:
```suggestion
the projectId in the tests has to match the project id in .firebaserc.
```

tests/README.md
```diff
@@ -0,0 +1,53 @@
...
+There doesn't seem to be a way to use the Emulator UI to inspect data from other
+projectIds.
+
+### Questions
+
+It isn't clear how Jest and firebase.assertSucceeds work together. The assertSucceeds
+return a promise. How does Jest know that the promise is complete? My best guess is
+that assetSucceeds has some Jest integration knowledge and internally calls Jest done
+or `expect.assertions(...)` with a expect statement.
```
kswenson:
```suggestion
or `expect.assertions(...)` with an expect statement.
```

kswenson (approved):
👍 LGTM -- I can't speak to whether the specific rules are correct, but this serves as a useful example of how to implement, document, and test Firebase rules for other projects to reference.

firestore.rules
```diff
@@ -15,62 +27,192 @@ service cloud.firestore {
...
+        // Make sure the learner isn't changing the current document's id
+        && request.resource.data.context_id == resource.data.context_id;
+    }
+
+    // This is similar to learnerOwner except that we are looking at the request.resource
+    // and in this case we can enforce the context_id matches the auth info
+    function learnerOwnerCreate() {
+      return request.auth.token.user_type == 'learner'
+        && string(request.auth.token.platform_user_id) == string(request.resource.data.platform_user_id)
+        && request.auth.token.platform_id == request.resource.data.platform_id
```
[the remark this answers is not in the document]
scytacki: Yeah I think so. There is a limit to the function stack but I think it is 10 so one nested function call should be fine.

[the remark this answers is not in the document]
scytacki: Oh actually looking at my own comment, we can't re-use it because in one case we are looking at `resource.data` and the other we are looking at `request.resource.data`.  It is possible to pass arguments to functions, so maybe we can parameterize it. I'll give it a shot...

dougmartin:
Looks good so far :+1:

scytacki:
@dougmartin this is ready now.

firestore.rules
```diff
@@ -15,62 +15,175 @@ service cloud.firestore {
...
+    }
+
+    function studentWorkRead() {
+      return anonymousRead()
+        || teacherOfContext()
+        || learnerOwner(resource);
+    }
+
+    function studentWorkUpdate() {
+      return anonymousUpdate()
```
kswenson, line 88: I would think that `notChangingIdentifyingProps()` would apply to the `anonymousUpdate()` as well.

dougmartin (approved):
Looks good :+1:

firestore.rules
scytacki: The concept applies, but that actual function is specific to the authenticated identifying props of platform_id and context_id. For anonymous the only identifying prop is the run_key, and also make sure they don't set a platform. So the anonymousUpdate checks those things itself.

PR #42, opened 2021-04-29
Upgrade automated s3 import

sfentress:
This extends @dougmartin 's original branch, adds some fixes for it, and then modifies it based on discussions and comments in the Google Doc.

The relevant section in the doc has been updated: https://docs.google.com/document/d/1F4ozfQOjzZfMMYdGiPp0MGWpAxjTbMa6lDIvPSnuNcI/edit#heading=h.qz7eu3tyf7n2

This PR is quite long. I'd be happy to do a synchronous code review at some point if that makes sense.


dougmartin (changes requested):
@sfentress I left some review questions - maybe it would be better to do this as a sync review with @scytacki and myself?

functions/src/auto-importer.ts
```diff
@@ -0,0 +1,263 @@
...
+
+1. createSyncDocAfterAnswerWritten runs on every write of an answer.  If the answer changes or is deleted a "sync doc"
+   is created or updated whose name is the run_key for the answer along with a boolean "updated".
+2. monitorSyncDocCount runs as a cron job every few minutes to find all the docs with updated = true.  Each document that is
+   found has a needs_sync field set to the current server time and has updated set to false.
+3. syncToS3AfterSyncDocWritten runs on every write of a sync doc.  If the needs_sync field exists and either the doc has
+   never synced before, or needs_sync > did_sync, it will gather up all the answers for that runKey and post them to
+   s3 as a parquet file. If there are no answers, it will delete the parquet file. It will then set a did_sync timestamp
+4. After the did_sync is set, syncToS3AfterSyncDocWritten will be notified again. If no new data has been written, it will
+   delete the sync doc
```
scytacki, line 39: If an answer is changed at the same time that syncToS3AfterSyncDocWritten is running, is it possible for it to delete the document which just had its 'updated' boolean set? 
In other words is it possible that the document fetched by syncToS3AfterSyncDocWritten could be changed before it is deleted?  

Perhaps the delete operation can be run in a transaction that includes the fetching of the document. I'd guess in that case the transaction will be aborted if the document changes. Or the operation to change the document will be blocked until the transaction has completed.

scripts/export-answers.js
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
 const BUCKET = "concordqa-report-data"
-const DIRECTORY = "partitioned-answers"
+const SOURCE = "authoring.staging.concord.org"
```
scytacki, line 4: I think our decision from earlier today is that this will look at all sources in Firestore. So it won't be necessary to keep track of this SOURCE.

scytacki (changes requested):
I think there might be a multi-thread issue when the sync document is deleted. Perhaps it can be solved by using a transaction.

scripts/export-answers.js
sfentress: Yes, I am looking into changing this now.

README.md
```diff
@@ -33,12 +40,27 @@ environment variables.  To create that file use:
 
 The `.runtimeconfig.json` file is present in the `.gitignore` so it won't be committed.
 
+### AWS secrets
+
+In order to write to S3, a user needs to be created in AWS IAM with permission to write to the S3 bucket.
+For instance there exists a user, `report-service-qa`, within the AminConcordQA role, with permission to write
+to the `concordqa-report-data` S3 bucket.
+
+We can download the AWS key and Secret Key for this user, and then push it up to Firebase with
```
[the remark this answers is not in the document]
sfentress: :+1:

scripts/export-answers.js
sfentress: This is now updated.

sfentress:
Changes since our last paired code-review:

* Parquet files are no longer saved in source subfolders in S3
* Sync_docs are no longer deleted
* Sync_docs now contain a bunch more logging information, including timestamps and running-time info
* Functions look across all sources in collection, so this will be useable as-is in production
* Bucket name is a firebase config variable
* Improved readme

scripts/export-answers.js
scytacki: So this SOURCE variable is just in the export-answers script and a developer will ned to set this appropriately before running the script.  We should make this a parameter, but that could be in a different PR which is focused on the export-answers script.

scytacki (approved):
This looks great to me.

scripts/export-answers.js
sfentress: Yes, I'm sorry, I missed the context of which script we were referring to when replying. In export-answers this is currently per-source. It could have been made into a parameter forcing a dev to specify it explicitly when running the script, or we could have made it work across all sources (though I think per-source makes sense when running the entire database dump), but it was not handled as part of this PR.

PR #43, opened 2021-05-04
Duplicate answer cleanup

eireland:


scripts/find-duplicate-answer.js
```diff
@@ -0,0 +1,85 @@
...
+
+const firestore = new Firestore();
+const answersRef = firestore.collection(`sources/authoring.staging.concord.org/answers`)
+const resourceRef = firestore.collection(`sources/authoring.staging.concord.org/resources`)
+let answerArr = [];
+let questionArr = [];
+let resourceArr = [];
+let questionUserMap = new Map;
+
+answersRef.orderBy("platform_user_id", "asc").orderBy("question_id", "asc").get().then(answerSnapshot => {
```
scytacki, line 23: This approach to avoid lots of memory usage is good.

But there needs to be a few changes to handle various conditions.
- The question_id is not guaranteed to be unique globally. Different activities might use the same question_id (in practice this isn't happening yet)
- The same student might have used this activity/sequence in multiple classes the logic above would treat their answers as duplicates even though in that case they are not.
- when an activity is run anonymously by LARA or the portal, the answers will not have a platform_user_id

Here is some logic that I think will work:
- if there is a resource_link_id then orderBy resource_link_id, platform_user_id, and question_id (and the check below on line 31 needs to be updated to check the resource_link_id too.
- if there is no resource_link_id this means the answer came from a anonymous run, so then orderBy run_key, and question_id

I'm not sure, but I'd guess that you'll add indexes to the database to do this kind of checking and ordering.

Also, I'd recommend running this against the production report-service-pro database. It has many more answers than staging, so it will probably quickly expose issues with memory.  For example I'd guess that the `answerSnapshot` returned by the query will be limited to some number of answers. So you will need to do some kind of loop requesting multiple answerSnapshots inorder to make your way through all of the answers. 

It isn't easy to describe, but if you look at this:
https://github.com/concord-consortium/report-service/blob/master/scripts/export-answers.js

It is using `startAfter` and `limit` to iterate over the huge amount of answers in production. Taking that approach will probably be necessary. It will probably also be necessary to add extra logic to figure out which document to `startAfter`. Since a group of duplicate answers might start just before the end of a batch and then continue into the next batch.

scytacki (changes requested):
The orderBy approach is good. There are several more issues that will need to be dealt with especially to handle the large number of answer documents on production.

PR #44, opened 2021-05-06
add new learner table fields to readme

mklewandowski:
Add the new learner fields to the readme.


scytacki (approved):
Looks good to me.

It is probably good to go ahead and delete the learners table (this doesn't delete the data) in both the concordqa account and production account and recreate it with this statement.

PR #45, opened 2021-05-07
Support Activity Player

sfentress:
This supports the Activity Player by using the LTI properties for logged in users that can be combined to identify a learner-assignment (platform_id, resource_link_id and platform_user_id) to find the answers collection and to create the parquet files. For anonymous users we still use the runKey.

For a logged-in user launching from the portal (LARA or Activity Player) we now save parquet files to:

    partitioned_answers/[escaped_resource_id]/[escaped-platform_id]/[escaped_resource_link_id]/[platform_user_id].parquet

For an anonymous user (LARA or AP):

    partitioned_answers/[escaped_resource_id]/anonymous/no-resource-link/[run-key].parquet

A few other changes needed to be made to support finding answers using these properties, and to support the differences in how the AP saves answers (some keys missing, some keys of different types, etc.)

Here are some examples of parquet files to the new very-wordy folder structure in S3:

Logged in user from LARA

https://s3.console.aws.amazon.com/s3/buckets/concordqa-report-data?region=us-east-1&prefix=partitioned-answers/https---authoring-staging-concord-org-sequences-71/https---learn-staging-concord-org/182/&showversions=false

Logged in user from AP:

https://s3.console.aws.amazon.com/s3/buckets/concordqa-report-data?region=us-east-1&prefix=partitioned-answers/https---activity-player-concord-org-branch-authenticate-and-fetch-data--activity-https-3--2--2-authoring-staging-concord-org-2-api-2-v1-2-activities-2-20753-json-firebase-app-report-service-dev-report-source-authoring-staging-concord-org/&showversions=false

Anonymous user from LARA:

https://s3.console.aws.amazon.com/s3/buckets/concordqa-report-data?region=us-east-1&prefix=partitioned-answers/https---authoring-staging-concord-org-activities-20960/&showversions=false

### DIFF
--- a/functions/src/auto-importer.ts
+++ b/functions/src/auto-importer.ts
@@ -68,7 +68,7 @@ interface SyncData {
 
 type PartialSyncData = Partial<SyncData>;
 
-const getHash = (data: any) => {
+export const getHash = (data: any) => {
   const hash = crypto.createHash('sha256');
   hash.update(JSON.stringify(data));
   return hash.digest('hex');
@@ -256,7 +256,7 @@ export const monitorSyncDocCount = functions.pubsub.schedule(monitorSyncDocSched
 });
 
 export const syncToS3AfterSyncDocWritten = functions.firestore
-  .document(`${answersSyncPathAllSources}/{id}`) // NOTE: {answerId} is a wildcard passed to Firebase
+  .document(`${answersSyncPathAllSources}/{id}`) // {id} is a wildcard passed to Firebase.
   .onWrite((change, context) => {
     return getSettings()
       .then(({ sync }) => {
--- a/functions/src/auto-importer.ts
+++ b/functions/src/auto-importer.ts
@@ -156,7 +156,13 @@ const syncToS3 = (answers: AnswerData[]): Promise<S3SyncInfo> => {
       await deleteFile();
       const writer = await parquet.ParquetWriter.openFile(schema, tmpFilePath);
       for (const answer of answers) {
+        // clean up answer objects for parquet
         answer.answer = JSON.stringify(answer.answer)
+        delete answer.report_state;
+        if (typeof answer.version === "number") {
+          answer.version = "" + answer.version;
+        }
+
         await writer.appendRow(answer);
       }
       await writer.close();
--- a/functions/src/auto-importer.ts
+++ b/functions/src/auto-importer.ts
@@ -14,22 +14,22 @@ const access = util.promisify(fs.access);
 const unlink = util.promisify(fs.unlink);
 const readFile = util.promisify(fs.readFile);
 
-import { AnswerData, schema, parquetInfo } from "./shared/s3-answers"
+import { AnswerData, schema, parquetInfo, AnswerMetadata, getAnswerMetadata, getSyncDocId } from "./shared/s3-answers"
 
 /*
 
 HOW THIS WORKS:
 
 1. createSyncDocAfterAnswerWritten runs on every write of an answer.  If the answer changes or is deleted a "sync doc"
-   is created or updated whose name is the run_key for the answer along with a boolean "updated". For logging purposes,
-   it will also add a last_answer_updated timestamp.
+   is created or updated whose name is a unique key for the learner-assignment (either a hash of the LTI data or a
+    runKey) along with a boolean "updated". For logging purposes, it will also add a last_answer_updated timestamp.
 2. monitorSyncDocCount runs as a cron job every few minutes to find all the docs with updated = true.  Each document that is
    found has a needs_sync field set to the current server time and has updated set to false.
 3. syncToS3AfterSyncDocWritten runs on every write of a sync doc.  If the needs_sync field exists and either the doc has
-   never synced before, or needs_sync > did_sync, it will gather up all the answers for that runKey and post them to
-   s3 as a parquet file. If there are no answers, it will delete the parquet file. It will then set a did_sync timestamp.
-   For logging purposes, it will also set a start_sync timestamp the momement it starts handling the sync_doc, and when
-   it finishes it will add some timing info.
+   never synced before, or needs_sync > did_sync, it will gather up all the answers for that learner-assignment and post
+   them to s3 as a parquet file. If there are no answers, it will delete the parquet file. It will then set a did_sync
+   timestamp. For logging purposes, it will also set a start_sync timestamp the momement it starts handling the sync_doc,
+   and when it finishes it will add some timing info.
 */
 
 const answerDirectory = "partitioned-answers"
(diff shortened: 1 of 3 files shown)


functions/src/auto-importer.ts
```diff
@@ -244,7 +256,7 @@ export const monitorSyncDocCount = functions.pubsub.schedule(monitorSyncDocSched
 });
 
 export const syncToS3AfterSyncDocWritten = functions.firestore
-  .document(`${answersSyncPathAllSources}/{runKey}`) // NOTE: {answerId} is a wildcard passed to Firebase
+  .document(`${answersSyncPathAllSources}/{id}`) // NOTE: {answerId} is a wildcard passed to Firebase
```
dougmartin, line 259: You might want to update this comment: `NOTE: {answerId} is a wildcard passed to Firebase`

dougmartin (approved):
Looks good 👍

functions/src/auto-importer.ts
```diff
@@ -255,15 +267,27 @@ export const syncToS3AfterSyncDocWritten = functions.firestore
           const needSyncMoreRecentThanStartSync = data.need_sync && (!data.start_sync || (data.need_sync > data.start_sync));
 
           if (data.need_sync && needSyncMoreRecentThanDidSync && needSyncMoreRecentThanStartSync) {
-            const syncDocRef = getAnswerSyncCollection(context.params.syncSource).doc(context.params.runKey);
+            const syncDocRef = change.after.ref;
```
scytacki, line 270: This is nice. I was wondering if you could optimize it like this.

functions/src/auto-importer.ts
```diff
@@ -285,7 +309,7 @@ export const syncToS3AfterSyncDocWritten = functions.firestore
                     .catch(functions.logger.error)
                 } else {
                   // if the learner has no answers associated with this run, delete the doc
-                  deleteFromS3(context.params.runKey, data.resource_url)
+                  deleteFromS3(data.answer_metadata)
```
scytacki, line 312: Should this be returned? And possibly `syncToS3` above? I'm wondering how onWrite handles promises that haven't resolved yet.

functions/src/shared/s3-answers.ts
```diff
@@ -1,51 +1,104 @@
...
-  type: { type: 'UTF8' },
+  type: { type: 'UTF8', optional: true },
   question_id: { type: 'UTF8' },
   source_key: { type: 'UTF8' },
-  question_type: { type: 'UTF8' },
-  tool_user_id: { type: 'UTF8' },
+  question_type: { type: 'UTF8', optional: true },
+  tool_user_id: { type: 'UTF8', optional: true },
   answer: { type: 'UTF8' },
   resource_url: { type: 'UTF8' },
```
scytacki, line 35: There will be some AP answer documents which have an incorrect resource_url. I fixed that here: https://github.com/concord-consortium/activity-player/pull/145 
I think we can just ignore those old answers for now, I think I got those fixed before the AP was used by real students. I'm just bring it up incase we can't find some data in the reports.

functions/src/shared/s3-answers.ts
```diff
@@ -1,51 +1,104 @@
...
   }
-  if (!runKey || !resourceUrl) {
-    throw Error(`Cannot create filename for ${runKey}`);
+}
+
+// returns just the part needed to identify the answer uniquely
+// urls don't need to be escaped because the document name will be hashed
+export const getSyncDocId = (answer: AnswerData) => {
+  if (answer.platform_id && answer.resource_link_id && answer.platform_user_id) {
+    return `${answer.platform_id}_${answer.resource_url}_${answer.platform_user_id}`;
```
scytacki, line 81: I'd guess that Firestore will have problems with some characters in its document ids. Since platform_id and resource_url are URLS I'd guess the `:` and `//` and possibly `?` will cause problems.

scytacki (changes requested):
Mostly looks good. I think I found an issue with the syncDocId and possibly a problem with how firestore's `onWrite` event handles promises. 

That second issue seems like it isn't a problem in practice, so perhaps it isn't a problem.

functions/src/auto-importer.ts
sfentress: Deeply buried in documentation....

functions/src/auto-importer.ts
sfentress: Sure.

I was looking and couldn't find a way to just say `.document(`${answersSyncPathAllSources}/*`)`, i.e. I don't care about the value of the wildcard. It seems like it might be a tiny bit more efficient, since I don't need it, but I think having something there is a necessity.

functions/src/shared/s3-answers.ts
sfentress: It could maybe be clearer in the code, but this string actually gets hashed before being turned into a firestore document id (in `addSyncDoc` in auto-importer.ts)

functions/src/auto-importer.ts
sfentress: Hmmm, we do the same with the call to `syncToS3` in the line just above this (307) and it appears to work fine (both in the successful case and the case where the write has some error).

The `onWrite` is always getting some return value, either the `getAllAnswersForLearner...` promise or `null`.

So the question is whether `getAllAnswersForLearner...then` should return the promise inside `then`. It's possible that you're right that it should, although fwiw Typescript doesn't actually like me returning those promises:

```
Argument of type '(querySnapshot: QuerySnapshot<DocumentData>) => Promise<void | WriteResult> | Promise<void | DeleteObjectCommandOutput>' is not assignable to parameter of type '(value: QuerySnapshot<DocumentData>) => void | WriteResult | PromiseLike<void | WriteResult>'.
```

This might be easily fixable by typing the promises somehow.

scytacki: Yeah I guess we should just leave it as is since it is working. 🤷 
My best guess is that if there was a problem we'd find some cases where did_sync was not set because firestore killed off the function before it finished writing to S3.  So if we find those kind of issues we might want to revisit this.

functions/src/shared/s3-answers.ts
```diff
@@ -1,51 +1,107 @@
...
   }
-  if (!runKey || !resourceUrl) {
-    throw Error(`Cannot create filename for ${runKey}`);
+}
+
+// returns just the part needed to identify the answer uniquely
+export const getSyncDocId = (answer: AnswerData) => {
+  if (answer.platform_id && answer.resource_link_id && answer.platform_user_id) {
+    // urls don't need to be escaped because the document name will be hashed
+    const ltiId = `${answer.platform_id}_${answer.resource_url}_${answer.platform_user_id}`;
```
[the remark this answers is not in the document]
sfentress: D'oh, good catch. :+1: That's a typo, I'm checking for `resource_link_id` in the if above.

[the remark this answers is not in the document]
scytacki: looks good.

PR #46, opened 2021-05-10
Support LTI in export-answers script

sfentress:
This updates the bulk-upload of all answers to support the LTI metadata and new directory structure of the update-answers Firestore function.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,27 @@
+# Running the export-answers script
+
+This script finds all student answers in a Firetore report-service database, collects all the answers for a single
+user's assignment into a single Parquet file, and uploads all the files to S3 for querying by the Athena database.
+
+```
+// build the shared library
+cd functions
+npm install
+npm run build:shared
+
+// build the scripts
+cd scripts
+npm install
+```
+
+Copy the `config.json.sample` file to `config.json` and fill in the AWS credentials for the S3 bucket. This should
+be available in 1Password as one of the report-service AWS accounts.
+
+Download the `credentials.json` file for the appropriate Firebase project from e.g.
+https://console.cloud.google.com/apis/credentials?project=report-service-pro&organizationId=264971417743
+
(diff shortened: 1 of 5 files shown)


scytacki (approved):
This looks good to me. 

It will likely make the script around 2 times slower. But I don't think there is anything simple we can do that about.  Sam and I discussed this in Slack. Here is my summary for the historical record:

The majority of the answers are currently from the LARA runtime, and these answers have both the LTI properties and the run_key property. When it is an anonymous answer the LTI properties are null. And when it is a portal answer it also has a valid run_key. So this means both queries will pick up all of the LARA answers both anonymous and launched from the portal. But we can't filter them out because of inconsistencies between the AP and LARA runtime answers.

The AP answers don't have the platform_id set at all for anonymous answers. And Firestore will skip documents that don't have a property set at all when you filter on it. So if we try to match documents with null platform_id's (that would be a way to just select LARA anonymous answers), that would then exclude all AP anonymous answers because they don't have a platform_id set at all. 

I think the right solution here is to make the answer documents more consistent and also make it easier for the report code to manage them without special casing the anonymous and portal based answers.

If all anonymous answers had a platform_id set to null, and resource_link_id set to null, and the platform_user_id set to the run_key.  Then we could just order all documents by those 3 fields. And the upload code would just need to know how to handle null values for those cases.

However that approach requires changes to both LARA and the AP, as well as a migration which updates all answers in firestore.  I'll make a separate story for this, so we don't forget this idea. In the meantime it seems best to go with this PR as is.

scripts/README.md
```diff
@@ -0,0 +1,27 @@
+# Running the export-answers script
+
+This script finds all student answers in a Firetore report-service database, collects all the answers for a single
```
kswenson, line 3:
```suggestion
This script finds all student answers in a Firestore report-service database, collects all the answers for a single
```

PR #52, opened 2021-05-14
Add question prompts to query

mklewandowski:
This PR adds question prompts to the Athena query generated in `generateSQL`.  Note that the prompts are only added to a single column for the multiple choice, image, and open response questions (image and open response are made up of more than one column).  A null value is added for additional columns since the UNION query requires an identical number of fields.

Example query with prompts:
```
-- activity-activity_20654

WITH activities AS ( SELECT * FROM "report-service"."activity_structure" WHERE structure_id = 'b1e8fe05-c62d-4d3c-b999-6bc61fa9c826' )

SELECT
  null as remote_endpoint,
  activities.questions['multiple_choice_12818'].prompt AS multiple_choice_12818_choice,
  activities.questions['multiple_choice_12843'].prompt AS multiple_choice_12843_choice,
  activities.questions['multiple_choice_12845'].prompt AS multiple_choice_12845_choice,
  activities.questions['multiple_choice_12846'].prompt AS multiple_choice_12846_choice,
  activities.questions['multiple_choice_12847'].prompt AS multiple_choice_12847_choice,
  activities.questions['multiple_choice_12848'].prompt AS multiple_choice_12848_choice
FROM activities

UNION ALL

SELECT
  remote_endpoint,
  activities.choices['multiple_choice_12818'][json_extract_scalar(kv1['multiple_choice_12818'], '$.choice_ids[0]')].content AS multiple_choice_12818_choice,
  activities.choices['multiple_choice_12843'][json_extract_scalar(kv1['multiple_choice_12843'], '$.choice_ids[0]')].content AS multiple_choice_12843_choice,
  activities.choices['multiple_choice_12845'][json_extract_scalar(kv1['multiple_choice_12845'], '$.choice_ids[0]')].content AS multiple_choice_12845_choice,
  activities.choices['multiple_choice_12846'][json_extract_scalar(kv1['multiple_choice_12846'], '$.choice_ids[0]')].content AS multiple_choice_12846_choice,
  activities.choices['multiple_choice_12847'][json_extract_scalar(kv1['multiple_choice_12847'], '$.choice_ids[0]')].content AS multiple_choice_12847_choice,
  activities.choices['multiple_choice_12848'][json_extract_scalar(kv1['multiple_choice_12848'], '$.choice_ids[0]')].content AS multiple_choice_12848_choice
FROM activities,
  ( SELECT l.run_remote_endpoint remote_endpoint, map_agg(a.question_id, a.answer) kv1, map_agg(a.question_id, a.submitted) submitted
    FROM "report-service"."partitioned_answers" a
    INNER JOIN "report-service"."learners" l
    ON (l.query_id = 'b1e8fe05-c62d-4d3c-b999-6bc61fa9c826' AND l.run_remote_endpoint = a.remote_endpoint)
    WHERE a.escaped_url = 'https---authoring-staging-concord-org-activities-20654'
    GROUP BY l.run_remote_endpoint )
```

image

### DIFF
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -61,23 +61,35 @@ exports.uploadDenormalizedResource = async (queryId, denormalizedResource, workg
 
 exports.generateSQL = (queryId, runnable, resource, denormalizedResource) => {
   const selectColumns = [];
+  const selectColumnPrompts = [];
   const escapedUrl = resource.url.replace(/[^a-z0-9]/g, "-");
 
   Object.keys(denormalizedResource.questions).forEach(questionId => {
     const type = questionId.split(/_\d+/).shift();
     switch (type) {
       case "image_question":
-        selectColumns.push(`json_extract_scalar(kv1['${questionId}'], '$.image_url') AS ${questionId}_image_url`)
-        selectColumns.push(`json_extract_scalar(kv1['${questionId}'], '$.text') AS ${questionId}_text`)
-        selectColumns.push(`kv1['${questionId}'] AS ${questionId}_answer`)
+        // add question prompt, include empty column because UNION query requires identical number of fields
+        selectColumnPrompts.push(`activities.questions['${questionId}'].prompt AS ${questionId}_image_url`);
+        selectColumnPrompts.push(`null AS ${questionId}_text`);
+        selectColumnPrompts.push(`null AS ${questionId}_answer`);
+
+        selectColumns.push(`json_extract_scalar(kv1['${questionId}'], '$.image_url') AS ${questionId}_image_url`);
+        selectColumns.push(`json_extract_scalar(kv1['${questionId}'], '$.text') AS ${questionId}_text`);
+        selectColumns.push(`kv1['${questionId}'] AS ${questionId}_answer`);
         break;
       case "open_response":
-        selectColumns.push(`kv1['${questionId}'] AS ${questionId}_text`)
+        // add question prompt, include empty column because UNION query requires identical number of fields
+        selectColumnPrompts.push(`activities.questions['${questionId}'].prompt AS ${questionId}_text`);
+        selectColumnPrompts.push(`null AS ${questionId}_submitted`);
+
+        selectColumns.push(`kv1['${questionId}'] AS ${questionId}_text`);
         // TODO: only add if can be submitted  (need to check resource structure)
-        selectColumns.push(`submitted['${questionId}'] AS ${questionId}_submitted`)
+        selectColumns.push(`submitted['${questionId}'] AS ${questionId}_submitted`);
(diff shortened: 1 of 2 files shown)


scytacki (changes requested):
The code changes look good to me.

Can you add a test for this?
I don't think there are any yet. I'd think the best test is a unit style test which passes in a set of arguments and checks the generate SQL.

So far there aren't any exclusive branches to test, so the whole thing can be tested with a single denormalizedResource object which contains all of the different types of questions.  The expected SQL in the test will serve as a nice way to document the SQL generated by the query-creator.

mklewandowski:
> The code changes look good to me.
> 
> Can you add a test for this?
> I don't think there are any yet. I'd think the best test is a unit style test which passes in a set of arguments and checks the generate SQL.
> 
> So far there aren't any exclusive branches to test, so the whole thing can be tested with a single denormalizedResource object which contains all of the different types of questions. The expected SQL in the test will serve as a nice way to document the SQL generated by the query-creator.

We do actually have a starting point for some tests in `report-service\query-creator\create-query\tests`.  I added a test in there, but it looks like the existing test wasn't working so I commented a few parts out for now.  Not really sure, however, if this is the right place.  Anyway, I added a simple test that confirms the query creation of a simple activity with an MC, open response, and image question.  This test succeeds locally when running `npm run test` inside `report-service\query-creator\create-query\`.  I'll push a commit momentarily and we can discuss further if this is what you had in mind.

scytacki:
@mklewandowski yes that test is exactly what I was thinking.

How do you run it locally?  

I'm wondering how hard it would be to set it up to run on github actions.

mklewandowski:
> @mklewandowski yes that test is exactly what I was thinking.
> 
> How do you run it locally?
> 
> I'm wondering how hard it would be to set it up to run on github actions.

In `report-service\query-creator\create-query` use `npm run test` to run the test locally.

Not sure how to connect it to gh actions...

scytacki:
OK I made a story for running the tests in GH Actions: https://www.pivotaltracker.com/story/show/178201558

scytacki (approved):

scytacki:
I'll merge this in. 

@mklewandowski and @sfentress  It'd be nice to demo this from learn.staging at the meeting today, but based on the readme I'm not confident that we can deploy an update to this without having to re-define the parameters setup for staging.

I made a story for documenting this better:
https://www.pivotaltracker.com/story/show/178201633

There are 2 issues which I documented in a comment in the story:
- will the existing parameters like the JwtHmacSecret be blown away when it is re-deployed?
- to support production and staging we need to specify two different buckets, so the samconfig.toml should have two environments to handle this.

Note these same concerns apply to this PR:
https://github.com/concord-consortium/rigse/pull/963#discussion_r622264611

PR #55, opened 2021-05-19
178184910 open response submitted column

mklewandowski:
This PR updates the Athena SQL query created in `generateSQL` so that a "submitted" column only appears for open response questions if they are marked as "required" (and hence show the submit button to the user).  When iterating through questions in the `denormalizedResource` object, we find the associated question in the `resource` object and check its `required` property.  This search should work for both a sequence and an activity.  The unit test for `generateSQL` has also been expanded to include required and not required open response questions.

Questions:
- are there other paths where we show the submit button?
- do we need to handle resources other than activity and sequence?

Generated query on activity with a required and a non-required open response question:
image

### DIFF
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -69,6 +69,24 @@ exports.generateSQL = (queryId, runnable, resource, denormalizedResource) => {
   ];
   const escapedUrl = resource.url.replace(/[^a-z0-9]/g, "-");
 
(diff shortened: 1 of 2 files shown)


scytacki (approved):
This looks great to me.
Only activities and sequences need to be handled for resource types.

I think there are other types of questions that can be required. But I'll make a new story for that since this one was specific to open-response questions.

PR #56, opened 2021-05-24
Fix getting answer metadata for new answers

sfentress:
Previously, we were using the previous version of the answer to get the answer's metadata. This was null if it was a new answer, so new answers were not getting sync docs created for them.


scytacki (approved):
looks good to me.

PR #57, opened 2021-05-24
Add `created` regex to export script

sfentress:
This adds the ability to filter the uploading of answers by a regex on the `created` property in an answer.

This is conservative: if an answer does not have a `created` field it will be uploaded, and it only takes one answer in a user's session that matches the regex (or has `created` missing) to upload the entire learner-assignment collection.

Also updated the readme with more information on running the script, including probably-pedantically-detailed Lightsail instructions.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -25,3 +25,52 @@ Run the script by passing in the destination S3 bucket and the source key for th
 ```
 node export-answers.js concordqa-report-data authoring.staging.concord.org
 ```
+
+the script accepts up to three arguments:
+
+```
+export-answers.js [s3-bucket] [source-key] [created-timestamp-regex]
+```
+
+* s3-bucket: the name of the s3 bucket we're going to put the answers in
+* source-key: the Firestore answers collection
+* created-timestamp-regex: (optional) a regex as a string, e.g. `2021-*`
+
+  if this is included we will gain a slight speed-up by not uploading answers whose `created` property does not match
+  the regex. Note that this is conservative: if an answer does not have a `created` field it will be uploaded, and it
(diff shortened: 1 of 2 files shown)


scripts/README.md
```diff
@@ -25,3 +25,52 @@ Run the script by passing in the destination S3 bucket and the source key for th
...
+
+the script accepts up to three arguments:
+
+```
+export-answers.js [s3-bucket] [source-key] [created-timestamp-regex]
+```
+
+* s3-bucket: the name of the s3 bucket we're going to put the answers in
+* source-key: the Firestore answers collection
+* created-timestamp-regex: (optional) a regex as a string, e.g. `2021-*`
```
scytacki, line 37: this example isn't a regex it is a glob

scripts/export-answers.js
```diff
@@ -9,6 +9,13 @@ const { parquetInfo, getSyncDocId, hasLTIMetadata, getHash, schema } = require('
 
 const BUCKET = process.argv[2];     // e.g. concordqa-report-data
 const SOURCE =  process.argv[3];    // e.g. authoring.staging.concord.org
+const DATE_REGEX  = process.argv[4] // optional, e.g. 2021-05*
```
scytacki, line 12: this example is a glob, as a regex it will match `2021-055555` it will also match anything that starts with `2021-0` so `2021-03` would also match.

scytacki (changes requested):
Looks good to me. It just has some documentation issues.

scripts/export-answers.js
sfentress: D'oh. Being a sleepy-head. You're right.

scytacki (approved):
looks good

PR #58, opened 2021-05-24
173668043 multiple choice query updates

mklewandowski:
This PR expands Athena SQL query for multiple choice questions to include:
- show all answers if question supports multiple answers
- show correct/wrong status if question has a valid answer defined

In doing so, the following types of multiple choice questions are now supported:
- multiple choice where user can enter single answer and correct answer is defined
- multiple choice where user can enter single answer and no correct answer is defined
- multiple choice where user can enter multiple answers and correct answer is defined
- multiple choice where user can enter multiple answers and no correct answer is defined

Unit tests have also been expanded to include these new cases. EDIT: I also used the `denormalizedResource` instead of the `resource` inside of `generateSQL` to access information about each question (since we have the `denormalizedResource` with the question structure, it felt redundant to pull that same structure from the `resource`).  I made one change to the `denormalizedResource` to allow this that is only used locally, but we should look and see if this affects anything else adversely.

The query itself is somewhat complicated, so here is an explanation of a few of the parts.  Here is an example of the portion of the query that for a given MC question, returns a user's answers with correct/wrong marked:
```
array_join(transform(CAST(json_extract(kv1['multiple_choice_13100'],'$.choice_ids') AS ARRAY(VARCHAR)), x -> CONCAT(activities.choices['multiple_choice_13100'][x].content, IF(activities.choices['multiple_choice_13100'][x].correct,' (correct)',' (wrong)'))),', ') AS multiple_choice_13100_choice
```

`array_join` will put all of the user answers for a given MC question into a single column.  However, we need to get the answers into an array first.

To do so we get the user answers (the `choice_ids`) in `kv1` as JSON with `json_extract`.  We the cast the JSON to an array (`CAST(...) AS ARRAY`).  We then use array `transform` to create a new array with the complete string for each answer - this includes if the answer is correct or wrong. 

However, not all questions have a correct answer.  To determine if a question has a correct answer we check the choices for a given question in the `denormalizedResource` to determine if any of the choices are marked correct.

If the question does have a correct answer, then we can mark each answer with `(correct)` or `(wrong)` by looking up the user answer in `activities.choices` and checking if it is correct:
```
IF(activities.choices['multiple_choice_13100'][x].correct,' (correct)',' (wrong)')
```

Here is an example of a full query that is generated (EDIT: updated from initial PR):
```
WITH activities AS ( SELECT *, cardinality(questions) as num_questions FROM "report-service"."activity_structure" WHERE structure_id = '888c7a74-35e4-4405-9dd3-bf1d876a577f' )

SELECT
  null as remote_endpoint,
  null as num_questions,
  null as num_answers,
  null as percent_complete,
  activities.questions['multiple_choice_13100'].prompt AS multiple_choice_13100_choice,
  activities.questions['multiple_choice_13103'].prompt AS multiple_choice_13103_choice,
  activities.questions['multiple_choice_13101'].prompt AS multiple_choice_13101_choice,
  activities.questions['multiple_choice_13102'].prompt AS multiple_choice_13102_choice
FROM activities

UNION ALL

SELECT
  remote_endpoint,
  activities.num_questions,
  cardinality(array_intersect(map_keys(kv1),map_keys(activities.questions))) as num_answers,
  round(100.0 * cardinality(array_intersect(map_keys(kv1),map_keys(activities.questions))) / activities.num_questions, 1) as percent_complete,
  array_join(transform(CAST(json_extract(kv1['multiple_choice_13100'],'$.choice_ids') AS ARRAY(VARCHAR)), x -> CONCAT(activities.choices['multiple_choice_13100'][x].content, IF(activities.choices['multiple_choice_13100'][x].correct,' (correct)',' (wrong)'))),', ') AS multiple_choice_13100_choice,
  array_join(transform(CAST(json_extract(kv1['multiple_choice_13103'],'$.choice_ids') AS ARRAY(VARCHAR)), x -> CONCAT(activities.choices['multiple_choice_13103'][x].content, '')),', ') AS multiple_choice_13103_choice,
  array_join(transform(CAST(json_extract(kv1['multiple_choice_13101'],'$.choice_ids') AS ARRAY(VARCHAR)), x -> CONCAT(activities.choices['multiple_choice_13101'][x].content, IF(activities.choices['multiple_choice_13101'][x].correct,' (correct)',' (wrong)'))),', ') AS multiple_choice_13101_choice,
  array_join(transform(CAST(json_extract(kv1['multiple_choice_13102'],'$.choice_ids') AS ARRAY(VARCHAR)), x -> CONCAT(activities.choices['multiple_choice_13102'][x].content, IF(activities.choices['multiple_choice_13102'][x].correct,' (correct)',' (wrong)'))),', ') AS multiple_choice_13102_choice
FROM activities,
  ( SELECT l.run_remote_endpoint remote_endpoint, map_agg(a.question_id, a.answer) kv1, map_agg(a.question_id, a.submitted) submitted
    FROM "report-service"."partitioned_answers" a
    INNER JOIN "report-service"."learners" l
    ON (l.query_id = '888c7a74-35e4-4405-9dd3-bf1d876a577f' AND l.run_remote_endpoint = a.remote_endpoint)
    WHERE a.escaped_url = 'https---authoring-staging-concord-org-activities-21014'
    GROUP BY l.run_remote_endpoint )
```

image

### DIFF
--- a/query-creator/create-query/steps/firebase.js
+++ b/query-creator/create-query/steps/firebase.js
@@ -79,7 +79,8 @@ const denormalizeActivity = (activity, denormalized) => {
     section.children.forEach(page => {
       page.children.forEach(question => {
         denormalized.questions[question.id] = {
-          prompt: question.prompt
+          prompt: question.prompt,
+          required: question.required ? question.required : false
         }
         if (question.type === "multiple_choice") {
           denormalized.choices[question.id] = {}
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -69,24 +69,6 @@ exports.generateSQL = (queryId, runnable, resource, denormalizedResource) => {
   ];
   const escapedUrl = resource.url.replace(/[^a-z0-9]/g, "-");
 
-  const questionIsRequired = (resourceObj, qId) => {
-    let isRequired = false;
-    if (resourceObj.type === "sequence") {
-      resourceObj.children.forEach((activity) => activity.children.forEach((section) => section.children.forEach((page) => page.children.forEach((question) => {
-        if (question.id === qId && question.required) {
-          isRequired = true;
-        }
-      }))));
-    } else if (resourceObj.type === "activity") {
-      resourceObj.children.forEach((section) => section.children.forEach((page) => page.children.forEach((question) => {
-        if (question.id === qId && question.required) {
-          isRequired = true;
-        }
-      })));
-    }
-    return isRequired;
-  }
-
   Object.keys(denormalizedResource.questions).forEach(questionId => {
     const type = questionId.split(/_\d+/).shift();
     switch (type) {
@@ -101,7 +83,7 @@ exports.generateSQL = (queryId, runnable, resource, denormalizedResource) => {
         selectColumns.push(`kv1['${questionId}'] AS ${questionId}_answer`);
         break;
       case "open_response":
-        let isRequired = questionIsRequired(resource, questionId);
+        const isRequired = denormalizedResource.questions[questionId].required;
 
         // add question prompt, include empty column because UNION query requires identical number of fields
         selectColumnPrompts.push(`activities.questions['${questionId}'].prompt AS ${questionId}_text`);
@@ -117,8 +99,15 @@ exports.generateSQL = (queryId, runnable, resource, denormalizedResource) => {
       case "multiple_choice":
         // add question prompt
         selectColumnPrompts.push(`activities.questions['${questionId}'].prompt AS ${questionId}_choice`);
-        const questionHasCorrectAnswer = `cardinality(map_filter(activities.choices['${questionId}'], (k, v) -> v.correct)) > 0`;
-        const answerScore = `IF(${questionHasCorrectAnswer}, IF(activities.choices['${questionId}'][x].correct,' (correct)',' (wrong)'), '')`;
+
+        let questionHasCorrectAnswer = false;
+        for (const choice in denormalizedResource.choices[questionId]) {
+          if (denormalizedResource.choices[questionId][choice].correct) {
+            questionHasCorrectAnswer = true;
+          }
+        }
+
+        const answerScore = questionHasCorrectAnswer ? `IF(activities.choices['${questionId}'][x].correct,' (correct)',' (wrong)')` : `''`;
         const choiceIdsAsArray = `CAST(json_extract(kv1['${questionId}'],'$.choice_ids') AS ARRAY(VARCHAR))`;
         selectColumns.push(`array_join(transform(${choiceIdsAsArray}, x -> CONCAT(activities.choices['${questionId}'][x].content, ${answerScore})),', ') AS ${questionId}_choice`);
         break;
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -117,7 +117,10 @@ exports.generateSQL = (queryId, runnable, resource, denormalizedResource) => {
       case "multiple_choice":
         // add question prompt
         selectColumnPrompts.push(`activities.questions['${questionId}'].prompt AS ${questionId}_choice`);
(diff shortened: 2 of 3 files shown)


mklewandowski:
Stepping back and looking at this leaves me with some questions about overall structure.  For some parts of the query generation, I'm not sure if it's better to do everything in the Athena SQL query or if some of this should be done by parsing the `resource` parameter passed into `generateSQL` and then using it to figure out if a question has, for example, a valid answer defined.  For example, I did this in a query to determine if there was an author defined correct answer:
```
cardinality(map_filter(activities.choices['${questionId}'], (k, v) -> v.correct)) > 0
```
But we also could have pulled this same information from the `resource` parameter using Javascript and then used that to construct the query.  @scytacki what do you think?  Should I try to get info out of the `resource` whenever possible or should I try to do as much of the work in the query itself when possible?

scytacki:
I think in this case it is probably better to get the info out of resource. I'm pretty sure this will make the query run faster. And hopefully it will also simplify the query text itself. I'd guess figuring out what the javascript resource processing is doing will be easier for developers as opposed to figuring out what the presto in the SQL statement is doing.

But as aside that presto statement is impressive. Nice job mastering this stuff. 👍

scytacki:
Another observation is that this makes me want to figure out how to make automated tests that actually run the SQL. That way we could verify that your refactoring to move the resource parsing to javascript produces the exact same result. 

Not part of this PR, though. :) I think it would require using the aws sdk to, uploading mock data to S3, running the query and checking the result.

mklewandowski:
> I think in this case it is probably better to get the info out of resource. I'm pretty sure this will make the query run faster. And hopefully it will also simplify the query text itself. I'd guess figuring out what the javascript resource processing is doing will be easier for developers as opposed to figuring out what the presto in the SQL statement is doing.
> 
> But as aside that presto statement is impressive. Nice job mastering this stuff. 👍

Yeah, I was thinking the same and leaning towards getting as much info as possible from the resource.  Seems there are several advantages:
- follows existing pattern of how we handle prompts by getting info from resource
- easier to debug javascript than the query result if we have errors or bugs
- allows more parts of the query generation to be tested in our unit tests
- probably makes query faster
- makes query easier to read and understand which in turn makes it easier to expand and modify 

I'm going to make some changes to this PR that lean on the resource when possible for the MC questions.  I can definitely determine if we have a valid authored correct answer by analyzing the resource.

mklewandowski:
@scytacki I made some changes to this PR, but there's still one piece that I want to discuss (and potentially change).
- I now determine if there is an author-specified correct answer by looking at the MC choices in the `denormalizedResource` inside of the `generateSQL` function rather than doing this in the query.  This simplifies a big chunk of the query.
- In the unit test rather than making both the `resource` and the `denormalizedResource` manually, I only make the `resource` and then call `firebase.denormalizeResource` to generate the `denormalizedResource`.  This should make the unit test better since it is now also testing our function that creates the `denormalizedResource`.
- Inside `generateSQL` I was getting information about the question (such as the `required` field from the `resource`.  However, we also have the `denormalizedResource`  available which is much easier to parse.  You'll notice that now I'm looking up if a question is required and if an MC question has an author-specified correct answer in the `denormalizedResource` .  That said, I DID make a change to the `denormalizedResource`  - I added the `required` field to the questions.  I'm not sure if this is safe or desired though.  This does eventually get uploaded as the activity structure so I'm not sure if I should be modifying or adding to the `denormalizedResource` .  I can easily go back to using the full `resource` to get the info that I need.  Thoughts?

query-creator/create-query/steps/firebase.js
```diff
@@ -79,7 +79,8 @@ const denormalizeActivity = (activity, denormalized) => {
     section.children.forEach(page => {
       page.children.forEach(question => {
         denormalized.questions[question.id] = {
-          prompt: question.prompt
+          prompt: question.prompt,
+          required: question.required ? question.required : false
```
mklewandowski, line 83: Here's the change to the `denormalizedResource` that I mentioned in my comment.  Not sure if this is safe.

scytacki (approved):
I think putting the required in the denormalized Resource is good. That will make it possible to use it in the Athena queries directly if we ever want to do that.

The simplified queries look good.

PR #60, opened 2021-05-27
Get learners with JWT

sfentress:
Use JWT to fetch and uploaded paginated learners

This updates the expected request from the portal to be using the JWT query, which then allows us to request learner details from the portal.

The function `aws.fetchAndUploadLearnerData` uses the JWT to request the leaners from the portal, and uploads each batch as it comes. (We no longer try to keep the learners in memory and upload them later, as we want to support requests of up to a million learners.)

Previously, each runnable got a queryId, and we uploaded the learners for that runnable to

`{s3_bucket}/learners/{queryId}/{queryId}.json`

Now, because our learners might be fetched in multiple parts, we upload to

```
{s3_bucket}/learners/{queryId}/{uuid-1}.json
{s3_bucket}/learners/{queryId}/{uuid-2}.json
...
```

We still keep one queryId per runnable.

As a side-effect, we no longer need an JwtHmacSecret. Requests from the portal no longer have the json signed, because they contain a signed JWT instead.

### DIFF
--- a/query-creator/create-query/app.js
+++ b/query-creator/create-query/app.js
@@ -1,5 +1,3 @@
-const { v4: uuidv4 } = require('uuid');
-
 const envVars = require("./steps/env-vars");
 const request = require("./steps/request");
 const firebase = require("./steps/firebase");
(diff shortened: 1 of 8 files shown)


query-creator/create-query/app.js
```diff
@@ -17,42 +15,35 @@ exports.lambdaHandler = async (event, context) => {
...
-    const json = request.validateJSON(body);
-    const runnables = request.getRunnables(json);
-    const user = json.user;
-
+    // get the post body
+    const body = request.getBody(event);
+    request.validateRequestBody(body);
+    const { json, jwt } = body;
+    const { query, learnersApiUrlxx, paginationSize, user } = json;
+    learnersApiUrl = "http://4d9f5154e993.ngrok.io/api/v1/report_learners_es/external_report_learners_from_jwt";
```
scytacki, line 26: I'd guess this should be removed?

query-creator/create-query/steps/request.js
```diff
@@ -1,106 +1,59 @@
...
+  }, (error) => {
+    throw error;
+  });
 }
 
-exports.getRunnables = (json) => {
+/**
+ * Returns [{runnableUrl, learners[]}, ...]
+ */
+exports.getLearnersPerRunnable = (learners) => {
```
scytacki, line 54: Maybe I miss understood, but I thought that this in memory loading of learners per runnable wasn't needed anymore?

scytacki (changes requested):
Looks good. I found the testing url was still included, and I think maybe some dead code?

query-creator/create-query/app.js
sfentress: d'oh, thanks

query-creator/create-query/steps/request.js
sfentress: For each batch of learners we receive, we still want to sort them by runnable_url so that we can upload them to the appropriate `queryId` for the appropriate runnable.

That is, unless we don't need this anymore, but I was assuming we're still splitting the query by runnable as we had been doing before.

So the batch of learners does need to be sorted in-memory, but we can forget about them for the next batch.

PR #61, opened 2021-05-28
Add support for activity player managed interactives in query creation

mklewandowski:
This PR adds the question type to the denormalizedResource which is then used to determine the question type when making the Athena query (rather than using the id).  This allows Activity Player managed interactive questions to work properly in queries.

### DIFF
--- a/query-creator/README.md
+++ b/query-creator/README.md
@@ -136,7 +136,7 @@ Next, you can use AWS Serverless Application Repository to deploy ready to use A
     ```
     CREATE EXTERNAL TABLE IF NOT EXISTS activity_structure (
       choices map<string,map<string,struct<content:string,correct:boolean>>>,
-      questions map<string,struct<prompt:string>>
+      questions map<string,struct<prompt:string,required:boolean,type:string>>
     )
     PARTITIONED BY
     (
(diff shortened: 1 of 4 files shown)


mklewandowski:
@scytacki at present, this PR and a previous PR added two fields to the denormalizedResource that is later updated as the activity structure.  These fields are not used by the query, however, and are only used in the JS code that makes the query.  Do you think we should update the activity structure?  The `question` field is what would need to change.  At present it is:
`questions map<string,struct<prompt:string>>`
And it would need to be something like:
`questions map<string,struct<prompt:string, required:boolean, type:string>>`

mklewandowski:
> @scytacki this PR and a previous PR added two fields to the denormalizedResource (`required` and `type`) - the denormalizedResource is later uploaded as the activity structure. These fields are not used by the query, however, and are only used in the JS code that makes the query. Do you think we should update the activity structure in the readme and in Athena? The `question` field is what would need to change. At present it is:
> `questions map<string,struct<prompt:string>>`
> And it would need to be something like:
> `questions map<string,struct<prompt:string, required:boolean, type:string>>`

I went ahead and updated the readme per our discussion this morning. This PR is ready to review.

scytacki (approved):
looks great

PR #63, opened 2021-06-02
Add results loader folder with starter project single page app as contents

mklewandowski:
This PR adds a new `results-loader` folder which contains the runnable contents of the CC `starter-projects` repo (a single page application).  This folder will be used as the starting point for the new results loader application.


scytacki (approved):
looks good

PR #65, opened 2021-06-04
Add secure run script for Mac users

mklewandowski:
This is a temporary solution to at least allow Mac devs to run the results-loader using https by running `npm run start:secure` and having the cert/key files in a `.localhost-ssl` folder inside their home directory.  Windows users need to manually change the path to the cert/key files (which works fine for me), but this should, for now, make it easier for devs running on Macs.  Eventually this should be replaced by the platform-independent solution that is on a branch in the `starter-projects`, but as discussed, that solution depends on a release of `webpack-dev-server`.  I think this is a good tweak to get in to make work on the results-loader a little easier.

### DIFF
--- a/results-loader/package.json
+++ b/results-loader/package.json
@@ -35,6 +35,8 @@
   },
   "scripts": {
     "start": "webpack serve",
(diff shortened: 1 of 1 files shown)


scytacki (approved):
looks good to me.

PR #66, opened 2021-06-07
Add portal authentication to results loader

mklewandowski:
This PR makes changes to the results loader to include portal authentication.  Changes include:
- check URL params for portal access token.  If one is not present, redirect to portal login
- display portal access token if available (temporary, will be removed as we keep building)
- if portal access token is obtained, request the firebase JWT
- display firebase JWT if available (temporary, will be removed as we keep building)
- adds additional layout and styling 

Obviously there's much more to do, but I wanted to start bringing work in so we can keep PRs small and so we have more options for working in parallel as we have more of the results loader in place.


results-loader/src/utils/results-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+    clientId: oauthClientName,
+    redirectUri: window.location.origin + window.location.pathname + window.location.search,
+    authorizationUri: `${portalUrl}${PORTAL_AUTH_PATH}`
+  });
+  // Redirect
+  window.location.href = portalAuth.token.getUri();
+};
+
+export const readPortalAccessToken = (portalUrl: string, oauthClientName: string): string => {
+  // No error handling to keep the code minimal.
```
scytacki, line 25: Maybe add a `TODO` comment here? 
I'm not sure what errors can happen at this point. The authorizeInPortal call is just constructing a URL to redirect to, so I don't think it can error. 
What is possible is that some error condition happens in the portal and it redirects back to the client with an error message.
This error redirect can happen here:
https://github.com/concord-consortium/rigse/blob/917d5aaa788502b00fe906749af80275203b83fd/rails/app/controllers/auth_controller.rb#L45-L47
The invalid response can happen for these 2 reasons:
https://github.com/concord-consortium/rigse/blob/917d5aaa788502b00fe906749af80275203b83fd/rails/app/models/access_grant.rb#L65-L74

Basically if the client is not configured correctly in the portal. Or the request from the client uses an invalid response type. The response type is handled by the ClientOAuth2 library so hopefully that would never be a problem. But if the client is not configured with the 'public' type then the portal should redirect back to the client with a `error` url param.

scytacki (changes requested):
Looks good.

One request to either handle the `error` parameter or add a TODO to handle this error case. Details are in a comment.

Another request is to add a TODO for removing the portal access_token from the URL after it is read. This way the user won't see it very long and it won't be easy for them to copy this URL and send it to someone else.

results-loader/src/utils/results-utils.ts
mklewandowski: For now I added these notes in a TODO comment.

mklewandowski:
> Looks good.
> 
> One request to either handle the `error` parameter or add a TODO to handle this error case. Details are in a comment.
> 
> Another request is to add a TODO for removing the portal access_token from the URL after it is read. This way the user won't see it very long and it won't be easy for them to copy this URL and send it to someone else.

I added your notes in a TODO comment.  

I also pulled the token out of the URL using a method from here:
https://stackoverflow.com/questions/22753052/remove-url-parameters-without-refreshing-page

results-loader/src/utils/results-utils.ts
mklewandowski: From looking at the above code that you linked, it appears that we will get an `error` URL param with a message that we can output to the user.  If I tweaked the return type of `readPortalAccessToken`, it could return an object that specifies either an error or a token.  Then we can output the error if we find one.

mklewandowski: I added one more commit that attempts to pull the portal error message from the URL and display that to the user.  So we have a first pass of handling errors from the portal that are communicated in the URL.

scytacki (approved):
The error handling and url updating is good. 👍 

I'll add a story about getting this deployed so we can start demoing it.

PR #68, opened 2021-06-08
Get athena workgroup resources and resource credentials

mklewandowski:
This PR gets the Athena workgroup resources based on the portal access token and firebase JWT.  It also gets the resource credentials for the first resource in the resource list.  Changes include:
- add `@concord-consortium/token-service`
- get list of athena workgroup resources and set 0th resource as the current resource
- display information about current resource in app
- get the credentials for the current resource and display the credential information in the app
- improve and unify the status messages at each step.  This makes it easier to follow the render code and easier to refactor chunks out into components if needed.

Note that when we call `listResources`, I'm setting the `tool` property to `example-app`.  Obviously this is wrong and need to be some string that is used by query-creator.  However, for now this makes debugging a little easier as these changes can be used in sync with resources created by the example-app.  I'm not entirely sure what the final value should be.

### DIFF
--- a/results-loader/src/utils/results-utils.ts
+++ b/results-loader/src/utils/results-utils.ts
@@ -1,4 +1,5 @@
 import ClientOAuth2 from "client-oauth2";
+import { TokenServiceClient, EnvironmentName, ResourceType, Resource, Credentials } from "@concord-consortium/token-service";
 
 const PORTAL_AUTH_PATH = "/auth/oauth_authorize";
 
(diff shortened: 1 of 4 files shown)


sfentress (approved):
This looks good to me. I'm still getting oriented in this app, but code seems fine and works locally.

results-loader/src/utils/results-utils.ts
```diff
@@ -43,3 +44,28 @@ export const getFirebaseJwt = (portalUrl: string, portalAccessToken: string, fir
...
+}
+export const getCredentials = async ({resource, firebaseJwt,
+  tokenServiceEnv}: IGetCredentials): Promise<Credentials> => {
+
+  // The jwt will be ignored if there is a readWriteToken on the resource
+  const client = new TokenServiceClient({ env: tokenServiceEnv, jwt: firebaseJwt });
+  const readWriteToken = client.getReadWriteToken(resource) || "";
+
+  return readWriteToken ? await client.getCredentials(resource.id, readWriteToken) :  await client.getCredentials(resource.id);
+};
```
scytacki, line 61: You can leave out the readWriteToken stuff here. The results-loader doesn't need to support them. 
Also I think an async function can return a promise without waiting for it, so the `await` can be dropped in the return statement. I'll be curious if typescript handles this correctly. If it does the return type of this function shouldn't need to be changed.
 
```suggestion
  return client.getCredentials(resource.id);
};
```

results-loader/src/utils/results-utils.ts
```diff
@@ -43,3 +44,28 @@ export const getFirebaseJwt = (portalUrl: string, portalAccessToken: string, fir
...
+
+export const listResources = async (firebaseJwt: string, amOwner: boolean,
+  tokenServiceEnv: EnvironmentName, resourceType: ResourceType) => {
+  const client = new TokenServiceClient({ jwt: firebaseJwt, env: tokenServiceEnv });
+  return client.listResources({
+    type: resourceType,
+    tool: "example-app", // TODO: this needs to be set to valid string based on the app that created the resource
```
scytacki, line 68: Right. Probably something like `"athena-reports"`.  But it is fine to leave this "example-app" for now.
When we switch to `"athena-reports"` we'll need to add a document in the token-service firestore database for this new tool.

scytacki (approved):
Looks good. Just one suggested change to remove the readWriteToken handling since it isn't needed.

results-loader/src/utils/results-utils.ts
mklewandowski: Fixed.

results-loader/src/utils/results-utils.ts
mklewandowski: Agreed.  In general we will need to lock down the results loader name in a handful of places and it will probably makes things easier if we are consistent (this tool name as well as places like the deployed URL, the oauth client name, etc.).  Then again, maybe it doesn't matter if they differ...

PR #71, opened 2021-06-09
Get query execution list and query execution, display to user

mklewandowski:
This PR gets the list of queries for the current workgroup and display information about these queries to the user in list form.  Changes include:
- request list of query execution ids for current workgroup
- for each query execution id, get query execution information
- using the query execution information, display the query date, query URI (won't be needed long-term, but is useful to show uniqueness for now), and present a button for downloading CSV (nothing is implemented here yet - that will be next).

### DIFF
--- a/results-loader/package-lock.json
+++ b/results-loader/package-lock.json
@@ -1314,6 +1314,11 @@
         "minimist": "^1.2.0"
       }
     },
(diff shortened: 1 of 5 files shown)


scytacki (approved):
Looks good to me.

PR #74, opened 2021-06-10
Report name update, Restore QueryItem component

mklewandowski:
This PR changes the name from "Results Loader" to "Researcher Reports" in a few more locations in the project and restores the QueryItem component that got lost in the last merge.


sfentress (approved):
Thank you. Sorry about that rebase mess-up

PR #76, opened 2021-06-11
173472187 add query download link

mklewandowski:
This PR adds a link to download the query results CSV file when the user presses the Generate CSV Link button.  The link is valid for 10 minutes.  Changes include:
- add `amazon-s3-uri` for parsing bucket, region, and key info from output location URI
- when button is pressed get info from output location URI and use info to make signed link
- show error messages if steps fail
- CSS tweak to better display app
- remove obsolete files

To test, you need to login to Athena and set the workgroup to the one that you created based on the info in the example app.  You then need to ensure that your query results are being stored in the `workgroup-output` folder in the `concordqa-report-data` bucket (not `s3://concordqa-report-data/athena-results/`).  Under settings, change the query result location to be: 
`s3://concordqa-report-data/workgroup-output/[WORKGROUP-NAME]`.  For example, in the workgroup below, I set it to `s3://concordqa-report-data/workgroup-output/test-resource-YT5DXLKCUhavZKVYQYnd/`.   

Once set, you can perform some queries in the Athena console and the results should be written to the `workgroup-output` folder in the `concordqa-report-data` bucket.  Those queries will appear in the researcher report and can be DLed.

image


researcher-reports/src/components/query-item.tsx
```diff
@@ -1,5 +1,7 @@
 import React, { useState, useEffect } from "react";
 import * as AWS from "aws-sdk";
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const AmazonS3URI = require("amazon-s3-uri");
```
scytacki, line 4: Did you try `import AmazonS3URI from "amazon-s3-uri";`

sfentress (approved):
LGTM

scytacki (approved):
Looks good to me.

researcher-reports/src/components/query-item.tsx
mklewandowski: Ah, I thought I tried that, but I must not have.  I must have been too hung up on the syntax that they presented in their example.  Fixed.

PR #79, opened 2021-06-15
Redirect to researcher reports

sfentress:
This adds the redirect to the researcher-reports app after the queries have been kicked off in Athena (unless `debugSQL=true` is in the query params).

The researcher-reports url is defined in an env variable set on deploy. This will not usually need to change for different portals, but it seemed helpful to have staging point to the master branch of the reports app.

When the query-creator loads, it extracts the portal url from the learner-api url that is passed in by the JWT. It then opens the researcher-reports with `?portal=purtalUrl`. The researcher reports app is updated to find this query param and use it for authentication with the portal. When the app later strips out the token, we preserve this param because we later need to make a request to the same portal for the JWT.

If the researcher-report is launched with no query param, it defaults to learn.portal.

### DIFF
--- a/query-creator/create-query/app.js
+++ b/query-creator/create-query/app.js
@@ -63,12 +63,23 @@ exports.lambdaHandler = async (event, context) => {
       }
     }
 
-    const message = sqlOutput.length ? sqlOutput.join("\n\n------\n\n") : "Success"
+    let message = sqlOutput.length ? sqlOutput.join("\n\n------\n\n") : "Success"
 
-    // TODO: redirect the user to the result loader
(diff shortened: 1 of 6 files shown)


sfentress:
One thing that this PR doesn't tackle yet is streaming any info to the learner while the queries are being created. This seemed low-priority, since we can only deal with 10,000 students at a time anyway, and a query of that size took ~16 seconds for me. While this is a little long, and it would be nice to give the user feedback that something is happening, it seemed non-critical.

scytacki (approved):
Looks good to me.

PR #81, opened 2021-06-16
Add completion status to the researcher report

mklewandowski:
This PR adds the query status to the researcher reports app.  The status is obtained from the `QueryExecution` object and displayed to the user for each query in the results list.

image


scytacki (approved):
looks good to me

PR #82, opened 2021-06-16
Researcher Report: Improve layout and text in report and status/debug sections

mklewandowski:
This PR improves the layout and text in the researcher report so it is more clear where the reports are and what the status/debugging information is for.  Changes include:
- move reports above debug/status info
- use "report" instead of "query" in user-facing text
- add border and title around report and debug/status sections to make them more clear

image


scytacki (approved):
looks good to me

PR #86, opened 2021-06-21
Use production token service on production

sfentress:
This tests to see if we're running on the production url, and, if so, use "production" tokenServiceEnv. Otherwise default to "staging".


scytacki (approved):
Looks fine to me.

We might need to allow overriding of the tokenServiceEnv for testing, but lets go with this until we have a use case for testing.

PR #89, opened 2021-06-29
use column abstraction

scytacki:
This way the columns for the header row and the columns for the main select should always line up.
This also makes the use of AS more consistent.
The tests had to be updated for this use of AS.

This should also pave the way for breaking the grouping sub query into two parts and moving
it to the WITH clause


mklewandowski (approved):
👍 This looks good to me.  I ran through all of my test cases in both modes and queries all come out as expected from what I can tell.

PR #90, opened 2021-06-29
Use new ES pagination API

sfentress:
This updates the API we use to get learners from the portal with the new pagination parameter.

This relies on getting concord-consortium/rigse#1038 merged in, and having the rake task run which updates all the learners with the new `created_at` field.


scytacki (approved):
looks good to me

PR #91, opened 2021-06-29
Add query support for learners with no answers

mklewandowski:
This PR adjust the Athena SQL query in `generateSQL` so that it can handle learners who we do not have answer data for.  This is based off of the query presented by @scytacki on Slack.  A few things that still merit testing or improving:
- in some of the queries there are excess line breaks (not sure if we care)
- probably should be tested more with real data before merging (the queries work, but I haven't set up the staging case yet where we have users who don't have answers so we can ensure that they appear in the report).

### DIFF
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -282,14 +282,6 @@ UNION ALL
 `;
     }
 
-    // const answerMaps = `map_agg(a.question_id, a.answer) kv1, map_agg(a.question_id, a.submitted) submitted`;
-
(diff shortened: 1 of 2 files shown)


scytacki:
I found an activity I had on staging which had 2 students that had run it before. Because the staging report-service was wiped out recently there were no answers recorded for these students. I had to force the activity to be republished because the resource structure was wiped out too. The portal learner researcher report filter page reported 2 learners. Then when using the staging athena report I got 0 learners.  When using a locally check out of this branch I got a query which resulted in showing the 2 learners. So I think this works.

scytacki (approved):
This works for me.
Thanks Matt!

PR #94, opened 2021-08-11
update rules to support new researcher jwts

scytacki:
These JWTs have a class_hash and a target_user_id.
The target_user_id should match the platform_user_id of the document being requested.
The class_hash should match the context_id.
This way the researcher is restricted to just the student's work in a specific class.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -33,6 +33,14 @@ service cloud.firestore {
         && request.auth.token.platform_id == res.data.platform_id;
     }
 
+    // Check for researchers: users with a class_hash and target_user_id
+    function researcherOfOwner(res) {
+      return request.auth.token.user_type == 'user'
+        && string(request.auth.token.target_user_id) == string(res.data.platform_user_id)
+        && request.auth.token.class_hash == res.data.context_id
+        && request.auth.token.platform_id == res.data.platform_id;
+    }
+
     function notChangingIdentifyingProps() {
       // Make sure the learner isn't changing the document's identifying properties
       return string(request.resource.data.platform_user_id) == string(request.auth.token.platform_user_id)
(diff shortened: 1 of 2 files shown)


scytacki:
This goes along with this PR: 
https://github.com/concord-consortium/rigse/pull/1062

The portal-report codebase still needs to be updated to send the target_user_id when requesting the JWT from the portal. Once that is done, then the full round trip of this approach can be tested.

scytacki:
It is difficult to do an integration test of these rules locally with a local portal and local portal-report.  I'm pretty confident this approach will work. 

So I'd like to try getting this merged to master and then deploy the rules to report-service-dev.

firestore.rules
```diff
@@ -55,18 +63,21 @@ service cloud.firestore {
...
 
     function anonymousCreate() {
-      return request.resource.data.run_key.size() > 10
+      return ('run_key' in request.resource.data)
+        && request.resource.data.run_key.size() > 10
         // Make sure anonymous answers don't have platform_ids
         && !('platform_id' in request.resource.data);
     }
 
     // Anyone can read any anonymous docs with a valid run_key
```
kswenson, line 72: This comment seems out of date.

pjanik (approved):
Looks good to me 👍

firestore.rules
scytacki: @kswenson Which comment are you referring to?

The one directly above `Anyone can read any anonymous docs with a valid run_key` is still valid. The only change I made was to make sure the resource.data (doc contents) has a run_key property before checking the size of the value of the property. The code actually worked before I think because firestore was automatically catching exceptions and treating them as false return values from functions.  But while testing with the emulator it seems to work differently.

kswenson: Unless I'm misreading it, the previous rule said that anyone with a runKey longer than 10 characters could read any document. With your change, anyone with a runKey longer than 10 characters _that matches the runKey of the document being requested_ can read it, which seems different than "anyone can read any anonymous document".

scytacki: The ` ('run_key' in request.resource.data)` means: "is there a key called 'run_key' in the object request.resource.data". So it isn't check check that the value of the run_key matches just that the key exists.

PR #96, opened 2021-08-14
researchers need access to feedback as well

scytacki:
The portal-report loads in the feedback even when just looking at a single answer
That code could be optimized to not request the feedback, but it also seems like it
would be useful if a researcher could open the student report and see the feedback given to the student.

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -1,5 +1,6 @@
 {
   "projects": {
-    "default": "report-service-dev"
+    "default": "report-service-dev",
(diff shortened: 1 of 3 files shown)


scytacki:
I've deployed these rules to report-service-dev already so I could test the full use case with a local portal, local lara, and local portal-report.

So hopefully we won't need more changes.

pjanik (approved):
Looks good 👍

PR #92, opened 2021-07-23
Add model urls to Athena query

sfentress:
This constructs a url to the Portal Report for every iframe interactive, if the user has answer data for the model.

The url that is constructed looks like:

```
https://portal-report.concord.org/branch/master/index.html
  ?auth-domain=[https://learn.staging.concord.org | https://learn.concord.org | other portal url]
  &firebase-app=[report-service-dev | report-service-pro]
  &iframeQuestionId=[model-id, eg. "mw_interactive_210741"]
  &class=[{portal-url}/api/v1/classes/{class-id}]
  &offering=[{portal-url}/api/v1/offerings/{offering-id}]
  &studentId=[student-id]
```

For this to work we need the `offering-id`, which wasn't previously passed back from the portal when requesting learner details, and has now been added as a PR. The leaners table must also be updated to support this.

An example query that can be run in Athena as `AdminConcordQA` is

```
-- id activity-activity_20811
-- name Sam's GeoCode test
-- type activity

WITH activities AS ( SELECT *, cardinality(questions) AS num_questions FROM "report-service"."activity_structure" WHERE structure_id = 'f6363e70-c366-4e02-82db-0b42fc9853f4' ),

grouped_answers AS ( SELECT l.run_remote_endpoint remote_endpoint, map_agg(a.question_id, a.answer) kv1, map_agg(a.question_id, a.submitted) submitted
  FROM "report-service"."partitioned_answers" a
  INNER JOIN "report-service"."learners" l
  ON (l.query_id = 'f6363e70-c366-4e02-82db-0b42fc9853f4' AND l.run_remote_endpoint = a.remote_endpoint)
  WHERE a.escaped_url = 'https---authoring-staging-concord-org-activities-20811'
  GROUP BY l.run_remote_endpoint ),

learners_and_answers AS ( SELECT run_remote_endpoint remote_endpoint, runnable_url, learner_id, student_id, user_id, offering_id, student_name, username, school, class, class_id, permission_forms, last_run, teachers, grouped_answers.kv1 kv1, grouped_answers.submitted submitted,
  IF (kv1 is null, 0, cardinality(array_intersect(map_keys(kv1),map_keys(activities.questions)))) num_answers
  FROM activities, "report-service"."learners" l
  LEFT JOIN grouped_answers
  ON l.run_remote_endpoint = grouped_answers.remote_endpoint
  WHERE l.query_id = 'f6363e70-c366-4e02-82db-0b42fc9853f4' )

SELECT
  null AS remote_endpoint,
[...]

PR #102, opened 2021-09-30
add sourceKey and answersSourceKey to report urls

scytacki:
The sourceKey parameter is necessary to support testing on staging servers which will have different source keys. Its value was already being passed into the query creator, so now it is just pass down into the `getColumnsForQuestion` function so it can be added to each URL.
 
The answersSourceKey is required to support Activity Player and LARA which have different answerSourceKeys and the portal-report does not try to guess the right one. The value of this param is available in each of the answers in the report-services Firebase. These answers are sync's to S3 so they can queried by Athena, and the Athena `partitioned_answers` table already had a `source_key` column that contains this value.  The SQL needed to be updated so the value of this source_key for each answer could be included in links for the answer. This was done with a 3rd aggregate map so the question id can be used to look up the source_key  for a particular student row in the output. 

This PR also increases some of the mocking needed for the first mocha test to be closer correct. The test still actually fails with a 500 response though. In the past it was printing out an exception but then the test was still passing. This exception was confusing. Now at least it doesn't print an exception. To finish this up we'd either have to mock out some methods in order to bypass network calls, or we'd have to mock the network calls better.

If you look at the changes SQL in query-creator/create-query/tests/unit/test-handler.js you can see the main thing this PR is doing.


dougmartin (approved):
Looks good :+1:

query-creator/create-query/steps/aws.js
```diff
@@ -156,13 +156,16 @@ const getColumnsForQuestion = (questionId, question, denormalizedResource, authD
...
         ` CAST(class_id AS VARCHAR), `,
         `'&offering=${encodeURIComponent(`${authDomain}/api/v1/offerings/`)}',`,
         ` CAST(offering_id AS VARCHAR), `,
         `'&studentId=',`,
-        ` CAST(user_id AS VARCHAR)`,
+        ` CAST(user_id AS VARCHAR), `,
+        `'&answersSourceKey=', `,
+        ` source_key['${questionId}']`,
         `)`
       ].join("")
```
scytacki, line 170: I tried before to clean this up the syntax here but couldn't find any better way to write this.  The confusing part is that we are essentially doing concatenation both in javascript and SQL. 

Perhaps the most clear approach is a multiline whitespace stripping function. 
Another option is a to use SQL concat for every parameter and value.  That would probably look the most clean, but it would also be the least efficient for Athena to process (unless we we made it complex enough to join strings sometimes with javascript and concat them other times.

One of my previous attempts is here: https://github.com/concord-consortium/report-service/pull/92#discussion_r677585189

PR #104, opened 2021-10-26
Added update of collaborator's answers when run with others [#179921104]

dougmartin:


scytacki (approved):
This looks good to me.

The only issue is that the security rules haven't been updated to restrict what people can set for the collaboration_owner_id. But that can be in a new PR.

scytacki:
@dougmartin feel free to update report-service staging with this new function code whenever it makes sense to you. Please add a tag for it first though. Tags for the report service functions start with a `report-service-` prefix like: https://github.com/concord-consortium/report-service/releases/tag/report-service-v1.1.0

PR #112, opened 2021-11-03
Fixed set/update of collaborator answer id

dougmartin:
Fix bug where collaborator answer id was the same as the leader of the collaboration leading to a runtime bug when the collaborator ran the activity later as their would be duplicate answer ids.

### DIFF
--- a/functions/package-lock.json
+++ b/functions/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "functions",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


scytacki (changes requested):
Instead of adding a unique id, can you just strip the old one, then FS will add a unique id on its own?

scytacki:
Also I think if you remove the id from the data being passed to the `set` during an update, it should also work. As far as I can tell Firestore will keep the id of the current document even if it isn't included in the data.

For example see the example here: https://googleapis.dev/nodejs/firestore/latest/DocumentReference.html#set 
It is setting the content of a document and it is not including the id.

dougmartin:
The id is also in the answer data, which is why I manually generated the key (it was just using the firebase id) before this fix.  The id inside the answer is not set by Firebase but rather by the activity player so I wanted to keep the data consistent.  Here is where the activity player sets it:

https://github.com/concord-consortium/activity-player/blob/436e96e20a47122df8288a1966b191f70cff2dd9/src/utilities/embeddable-utils.ts#L106

scytacki (approved):
OK, I see that now.

I'm not sure why we are doing that, but oh well. 🤷

PR #116, opened 2021-12-06
180294630 athena learner log queries 

knowuh:
This PR adds a new Learner Log report Athena Query.  The Learner report query is triggered when the `useLogs` query parameter is sent to the report service.  It includes other minor changes:

- Fix deployment action issues (Pin node to 14)
- Add some documentation for updating deployment parameters.
- Use Log Query when query param `useLogs` present
- Update documentation for query parameters
- Add query polling in researcher reports
- Change time format on query display in researcher reports


scytacki (approved):
Looks good to me. 👍

PR #118, opened 2021-12-09
180559584 Athena Log queries are decoupled from learner information.

knowuh:
## Athena Log queries can be decoupled from learner information:

The goal of this PR is to be able to run learner log queries optionally leaving out learner meta data.

1. 56b9800 : 2021-12-09 :  Update instructions for developing with local docker using ngrok public proxy.

[PT Story #180559584](https://www.pivotaltracker.com/story/show/180559584)


scytacki (approved):
seems good to me

PR #119, opened 2021-12-17
180559584 -- limit CSV columns when reporting on log data only.

knowuh:
- Rename query param `endpointOnly` to `narrowLearners`. It makes sense for the portal learner endpoint to use `endpoint_only` for its param, but here that param name seems vague. Renamed for clarity, and updated the docs.

- If the query param `narrowLearners` is true, The select statement will only include `log.*` columns, because we didn't run the reconsituting learner queries. This just reduces empty values in the CSV file that is created.

[#180559584]

### DIFF
--- a/query-creator/create-query/app.js
+++ b/query-creator/create-query/app.js
@@ -19,7 +19,7 @@ exports.lambdaHandler = async (event, context) => {
     const tokenServiceEnv = params.tokenServiceEnv;
     const usageReport = params.usageReport || false;
     const useLogs = params.useLogs || false;
-    const endpointOnly = params.endpointOnly || false
+    const narrowLearners = params.narrowLearners || false
 
     if (!reportServiceSource) {
       throw new Error("Missing reportServiceSource in the report url");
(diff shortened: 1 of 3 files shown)


query-creator/create-query/app.js
```diff
@@ -19,7 +19,7 @@ exports.lambdaHandler = async (event, context) => {
     const tokenServiceEnv = params.tokenServiceEnv;
     const usageReport = params.usageReport || false;
     const useLogs = params.useLogs || false;
-    const endpointOnly = params.endpointOnly || false
+    const narrowLearners = params.narrowLearners || false
```
kswenson, line 22: Briefer option for accomplishing the same thing:
```typescript
  const { reportServiceSource, debugSQL = false, tokenServiceEnv, usageReport = false, useLogs = false, narrowLearners = false } = params;
```

knowuh:
@scytacki -- Things were working fine on staging, but I realized I didn't like two things: 

1. The query param name 'endpointOnly' wasn't very discriptive.
2. The CSV file was coming back with a lot of extra null values for the learner in the case when we were not doing a full join for learner data, or doing extra portal queries. I changed the select statement to omit the learner meta data.

scytacki (approved):
looks goood to me

query-creator/create-query/app.js
scytacki: the more concise version would behave slightly differently: if one of these params was the empty string `""` the current code would turn that into false. The concise version would leave the value `""`. 

I don't know if that would matter in this case.

I'm going to merge this as is, but feel free to make it more concise if it makes sense.

PR #123, opened 2022-01-19
180969564 : Query-Creator: Database Name param

knowuh:
Query-Creator:
The Log database name changes from staging to production.

We need to put the DB name in an ENV var to be configured in cloud formation.

[#180969564]
https://www.pivotaltracker.com/story/show/180969564

(FWIW this has already been deployed to the production query-creator stack on AWS)


scytacki (approved):
Looks good to me.

PR #159, opened 2022-10-26
Added user raw log handling [#180951212]

dougmartin:
This adds support for the user report request from the portal.  It also adds a tsconfig.json file to enable VSCode to type check the Javascript - it is not used to compile the code.

### DIFF
--- a/query-creator/create-query/app.js
+++ b/query-creator/create-query/app.js
@@ -6,7 +6,116 @@ const tokenService = require("./steps/token-service");
 
 const portalToAuthDomainMap = {
   "https://learn-report.staging.concord.org": "https://learn.staging.concord.org",
-  "https://learn-report.concord.org": "https://learn.concord.org"
+  "https://learn-report.concord.org": "https://learn.concord.org",
+}
+
+const trimTrailingSlash = (s) => s.trim().replace(/\/$/, "");
+
+const ensureWorkgroup = async (portalUrl, jwt, tokenServiceEnv, email) => {
+  const tokenServiceJwt = await request.getTokenServiceJwt(portalUrl, jwt);
+  const resource = await tokenService.findOrCreateResource(tokenServiceJwt, tokenServiceEnv, email, portalUrl);
+  const workgroupName = await aws.ensureWorkgroup(resource, email);
+
+  return workgroupName;
+}
+
+const userReport = async (body, tokenServiceEnv, debugSQL) => {
+  const { json, portal_token } = body;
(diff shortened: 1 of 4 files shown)


dougmartin:
@scytacki / @lublagg I've tagged you both for review.

The diff is a little messy - I left the common code in the main function and then split out the old learner and new user report code into two functions.

query-creator/create-query/steps/aws.js
```diff
@@ -361,6 +368,36 @@ exports.generateLogSQL = (queryId, runnableUrl, authDomain, sourceKey) => {
...
 };
 
+/*
+Generates a very wide row including all fields from the user log.
+*/
+exports.generateUserLogSQL = (usernames, activities, start_date, end_date) => {
+  const logDb = process.env.LOG_ATHENA_DB_NAME;
+  const where = [];
+  if (usernames.length > 0) {
+    where.push(`log.username IN (${usernames.map(u => `'${escapeSingleQuote(u)}'`).join(", ")})`)
```
scytacki, line 378: usernames might be large enough that we have to handle them like learner: upload json to S3 and do a join with it.

But I think it is good to start this more simple way and we can see if we run into problems.

query-creator/create-query/app.js
```diff
@@ -34,75 +140,7 @@ exports.lambdaHandler = async (event, context) => {
...
-
-    for (const runnableUrl in queryIdsPerRunnable) {
-      if(useLogs) {
-        await doLearnerLogReporting(runnableUrl);
-      }
-      else {
-        await doLearnerAnswerReporting(runnableUrl);
-      }
-    }
+    const {sqlOutput, portalUrl} = body.json.type === "users" ? await userReport(body, tokenServiceEnv, debugSQL) : await learnersReport(params, body, tokenServiceEnv, debugSQL, reportServiceSource);
```
scytacki, line 143: This is a long line. How about this formating:

```suggestion
    const {sqlOutput, portalUrl} = body.json.type === "users" 
      ? await userReport(body, tokenServiceEnv, debugSQL) 
      : await learnersReport(params, body, tokenServiceEnv, debugSQL, reportServiceSource);
```

scytacki (approved):
This looks good to me. I just had one formatting suggestion, and one note about possible future work.

query-creator/create-query/app.js
dougmartin: Done.  Updated in commit: https://github.com/concord-consortium/report-service/pull/159/commits/910cc6a8ca2b1a81e2b2a5e96b3574e171f9f7e5

PR #163, opened 2022-11-02
[#183618503]: Strip html from question prompts.

lublagg:


scytacki (changes requested):
Please add a test for this. There are some tests in test-hander.js they are already calling denormalResource with a mock resource. So I think it should be staightforward to add a new one that just tests the result of that function when there is html in the prompt.

scytacki (approved):
Nice. Looks good to me.

PR #164, opened 2022-11-03
[#183617855]: Multiple choice answers shown in second header.

lublagg:


scytacki (changes requested):
This looks good. I'm not totally sure but I think you'll have to update the table definition of the `activity_structure` table in order to extract the correct answer out of the question.

That schema is documented here: https://github.dev/concord-consortium/report-service/blob/c6808cfdeeda2707fe73d0b0f4a00c8102dbb44f/query-creator/README.md#L227-L231

Were you able to run this query in Athena? I would guess it would break without the table definition change. I thin there is some documentation somewhere that explains how to update the table definition.

query-creator/create-query/steps/aws.js
```diff
@@ -301,20 +304,35 @@ learners_and_answers AS ( SELECT run_remote_endpoint remote_endpoint, runnable_u
       });
       allColumns.push(...questionsColumns);
 
+      let columnWithAnswerText = allColumns.find(column => column.name.includes("_text"));
+      orderByText += columnWithAnswerText ? `\nORDER BY learner_id NULLS FIRST, ${columnWithAnswerText.name}`: "";
+
```
scytacki, line 309: I think this will have a problem when Doug's multiple activities report is included.
In that case there will need to be multiple learner_id fields for each "student+class" row. 
If a student hasn't done one of the activities then that learner_id field will probably be null.

Perhaps you can use the `class` column instead. Those should always be non-null.

If there are no open response questions in the activity then ordering seems like it will be messed up. You could populate the first column (currently remote_endpoint) with a fixed string like "correct answer" for this row. And then perhaps put "prompt" in that column for the prompts row. Now could could use that column as the secondary ordering instead of `columnWithAnswerText`. You'll have to use `DESC` since `correct answer` is before `prompt`.

scytacki (changes requested):
This ordering approach seems like what we'll have to do. 
I think I found a few problems with the columns you chose and suggested some other ones that should work out better.

scytacki (approved):
Looks good to me. 👍

PR #166, opened 2022-11-10
Fixed user report for AP logged activities and sequences [#183773943]

dougmartin:
The Activity Player logs the value of its sequence or activity query parameter in the activity slot in the log message.  This code checks for those parameters when mapping over the activities to query.


dougmartin:
@scytacki Due to not being able to create new activities on staging I tested the initial change on a CLUE activity with its teacher dashboard as I could directly create that on staging without publishing.  I only saw this issue when I tried to add the user report to production and tested against a published activity.

PR #168, opened 2022-11-15
[#178476237]: Add fields for required questions and answers in Learner Reports

lublagg:
After some discussion with Kiley a different approach was taken for this story. New criteria can be found here: https://www.pivotaltracker.com/n/projects/2441249/stories/178476237.

### DIFF
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -273,7 +273,9 @@ grouped_answers AS ( SELECT l.run_remote_endpoint remote_endpoint, ${answerMaps}
 
   const learnersAndAnswers = hasResource ? `
 learners_and_answers AS ( SELECT run_remote_endpoint remote_endpoint, runnable_url, learner_id, student_id, user_id, offering_id, student_name, username, school, class, class_id, permission_forms, last_run, teachers, grouped_answers.kv1 kv1, grouped_answers.submitted submitted, grouped_answers.source_key source_key,
-  IF (kv1 is null, 0, cardinality(array_intersect(map_keys(kv1),map_keys(activities.questions)))) num_answers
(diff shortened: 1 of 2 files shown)


query-creator/create-query/steps/aws.js
```diff
@@ -273,7 +273,9 @@ grouped_answers AS ( SELECT l.run_remote_endpoint remote_endpoint, ${answerMaps}
 
   const learnersAndAnswers = hasResource ? `
 learners_and_answers AS ( SELECT run_remote_endpoint remote_endpoint, runnable_url, learner_id, student_id, user_id, offering_id, student_name, username, school, class, class_id, permission_forms, last_run, teachers, grouped_answers.kv1 kv1, grouped_answers.submitted submitted, grouped_answers.source_key source_key,
-  IF (kv1 is null, 0, cardinality(array_intersect(map_keys(kv1),map_keys(activities.questions)))) num_answers
+  IF (kv1 is null, 0, cardinality(array_intersect(map_keys(kv1),map_keys(activities.questions)))) num_answers,
+  cardinality(filter(map_keys(activities.questions), x->activities.questions[x].required=TRUE)) num_required_questions,
+  IF (kv1 is null, 0, cardinality(filter(map_keys(submitted), x->submitted[x]=TRUE))) num_required_answers
```
scytacki, line 278:
```suggestion
  IF (submitted is null, 0, cardinality(filter(map_keys(submitted), x->submitted[x]=TRUE))) num_required_answers
```

Probably `submitted` would only be null if `kv1` is also null, but it seems better to be explicit about checking the map that is being used.

scytacki (changes requested):
Looks good. I just found one thing that should probably be changed.

Not required, but you could probably make this more efficient by using map_values instead of map_keys.

scytacki (changes requested):
I might have it wrong but I think you changed one of the filter functions without also changing the thing being filtered to map_values

scytacki (approved):
Looks good to me.

PR #169, opened 2022-11-17
Added question number to details report prompts [#183618540]

dougmartin:
This also adds a default prompt of "(no prompt)" and shows the prompt in the column headers of all the columns for the answer instead of just the first column.


scytacki (approved):
The addition of the question number to the prompt looks good.

As mentioned in slack, I don't think it is worth adding the prompt to each of the extra columns for a question. This can make these columns (and the whole table) much wider than necessary.

Kiley agrees at least in the case of the submitted column. 
It isn't a huge deal though, so if you'd rather just leave this and move on that is fine. We can wait to see what users think about it.

PR #170, opened 2022-11-23
Remove promptheader from submitted column.

lublagg:
Having the promptHeader value on the submitted columns was causing queries to fail because of the incompatibility between varchar and boolean values.


dougmartin (approved):
Looks good :+1:

PR #178, opened 2022-12-09
Multiple resource usage report & details report [#178474999]

dougmartin:


scytacki (changes requested):
As noted in Slack, there should be just one row for each unique `(user, class)`. In this current PR there is a new row for each learner. 

This gist provides an example SQL query that generates a single row for each unique `user, class)`: https://gist.github.com/scytacki/19c7d0e1ce8199409b5da9a4bbf3fb3d

query-creator/create-query/tests/unit/test-handler.js
```diff
@@ -115,280 +139,442 @@ describe('Denormalize resource', function () {
...
+          activities_1.questions['managed_interactive_66666'].prompt AS res_1_managed_interactive_66666_choice,
+          activities_1.questions['managed_interactive_77777'].prompt AS res_1_managed_interactive_77777_image_url,
+          activities_1.questions['managed_interactive_77777'].prompt AS res_1_managed_interactive_77777_text,
+          activities_1.questions['managed_interactive_77777'].prompt AS res_1_managed_interactive_77777_answer,
+          activities_1.questions['managed_interactive_88888'].prompt AS res_1_managed_interactive_88888_json,
+          activities_1.questions['managed_interactive_88888'].prompt AS res_1_managed_interactive_88888_url,
+          activities_1.questions['managed_interactive_99999'].prompt AS res_1_managed_interactive_99999_json,
+          activities_2.questions['managed_interactive_88888'].prompt AS res_2_managed_interactive_88888_json,
+          activities_2.questions['managed_interactive_88888'].prompt AS res_2_managed_interactive_88888_url
+        FROM activities_1, activities_2
```
scytacki, line 275: I think this from will have the same problems. When activities_1 is empty then the result of this query will be empty.

query-creator/create-query/tests/unit/test-handler.js
```diff
@@ -115,280 +139,442 @@ describe('Denormalize resource', function () {
...
+          activities_1.questions['managed_interactive_66666'].correctAnswer AS res_1_managed_interactive_66666_choice,
+          null AS res_1_managed_interactive_77777_image_url,
+          null AS res_1_managed_interactive_77777_text,
+          null AS res_1_managed_interactive_77777_answer,
+          null AS res_1_managed_interactive_88888_json,
+          null AS res_1_managed_interactive_88888_url,
+          null AS res_1_managed_interactive_99999_json,
+          null AS res_2_managed_interactive_88888_json,
+          null AS res_2_managed_interactive_88888_url
+        FROM activities_1, activities_2
```
scytacki, line 338: I think this from will have the same problems. When activities_1 is empty then the result of this query will be empty.

scytacki (changes requested):
I found two more places where the query won't work if the activities_x virtual table has no rows. 

I'm not sure what kind of join to use here. Perhaps an `full outer join` with a `1=1` condition. Since the activities table will have either 0 or 1 rows I think this will always result in a 1 row result.

query-creator/create-query/tests/unit/test-handler.js
dougmartin: Fixed in https://github.com/concord-consortium/report-service/pull/178/commits/036a56584b621c5049c13ea4d0e47a031c5f91bd

query-creator/create-query/tests/unit/test-handler.js
dougmartin: Also fixed in https://github.com/concord-consortium/report-service/pull/178/commits/036a56584b621c5049c13ea4d0e47a031c5f91bd

scytacki (approved):
looks good.

PR #187, opened 2023-01-11
Fixed time based user log queries [#184157275]

dougmartin:
The time based query was using the old Postgres string date format whereas Athena needed a unix timestamp.


lublagg (approved):
Looks good 👍

PR #196, opened 2023-10-12
fix: Filter out null runnable urls [PT-185994904]

dougmartin:
This changes the !== undefined runnable url filter to non-falsy so that runnable urls with null values are also filtered.

**NOTE**: this PR also updates the test matrix to use node 16 based on a required upgrade message from the Firebase CLI which when cased node-sass to break.  The node-sass package isn't being used in this project so I removed it instead of converting it to use dart-sass.


lublagg (approved):
Looks good to me 👍

PR #197, opened 2023-10-13
fix:  JSON definition showing up instead of answer [PT-185994915]

dougmartin:
When there is no answer to a question interactive open response question the report state is saved as the answer in the report hash.  This change detects if the open response answer looks like the report state, and if so, returns an empty string as the answer instead of the JSON.


lublagg (approved):
Nice. Looks good 👍

PR #198, opened 2023-10-13
fix: Update report completion status and generate csv button [PT-185257065]

dougmartin:
This updates the completion status to use different text colors depending on the status.  It alss disables the generate csv button until the completion status is succeeded.

### DIFF
--- a/researcher-reports/src/components/query-item.tsx
+++ b/researcher-reports/src/components/query-item.tsx
@@ -126,6 +126,21 @@ export const QueryItem: React.FC<IProps> = (props) => {
     }
   };
 
+  const lowerQueryCompletionStatus = queryCompletionStatus.toLowerCase();
+  const running = lowerQueryCompletionStatus === "running";
+  const succeeded = lowerQueryCompletionStatus === "succeeded";
+  const failed = lowerQueryCompletionStatus === "failed";
+
+  // show the generate button until it succeeds and it is clicked
+  // the button will be disabled until it succeeds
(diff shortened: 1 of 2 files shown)


researcher-reports/src/components/query-item.tsx
```diff
@@ -126,6 +126,21 @@ export const QueryItem: React.FC<IProps> = (props) => {
...
+  const succeeded = lowerQueryCompletionStatus === "succeeded";
+  const failed = lowerQueryCompletionStatus === "failed";
+
+  // show the generate button until it succeeds and it is clicked
+  // the button will be disabled until it succeeds
+  const showGenerateCSVLinkButton = !succeeded || !downloadURL;
+
+  const renderCompletionStatus = () => {
+    const suffix = running ? "... (please wait)" : "";
+    const className = running ? "running" : (succeeded ? "succeeded" : (failed ? "failed" : ""));
```
lublagg, line 140: Just curious, is there a reason not to just use lowerQueryCompletionStatus as the className?

lublagg (approved):
Looks good 👍  just had a small question about one line

researcher-reports/src/components/query-item.tsx
dougmartin: Good point :)  There isn't really a reason now that the display code converged on that.  I'll change it - thanks!

dougmartin: @lublagg fixed in https://github.com/concord-consortium/report-service/pull/198/commits/1e2eb0868355f1f815847644a96c77eef905a993

PR #200, opened 2023-10-17
fixed: Interpolation dollar signs showing around completion status

dougmartin:
This was a bug introduced during a PR review when the completion status display code was refactored.


scytacki (approved):
looks fine. :)

PR #207, opened 2024-03-13
feat: Add url column for open response questions [PT-181870149]

dougmartin:
This adds an url column to each open response question that points to the portal dashboard so that researchers can listen to audio responses.


pjanik (approved):
Looks good 👍

PR #208, opened 2024-03-13
build: Update ci actions to node 18

dougmartin:


dougmartin:
Merging without review - just a config change for the test runner.

PR #209, opened 2024-03-20
fix: Always add portal report url for open response [PT-181870149]

dougmartin:
If the student only answers with audio the answer text looks the same as if they didn't enter text so we always need to add the url instead of making it dependent on if the text was updated.


dougmartin:
Merging without review.

PR #210, opened 2024-03-27
feat: Add research read access to classes [PT-187154204]

dougmartin:
NOTE: this also adds a missing ap_runs matcher that exists on both staging and production and must have been added there via the web UI.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -25,8 +25,12 @@ service cloud.firestore {
         && request.auth.token.platform_id == resource.data.platform_id;
     }
 
-    // Check if answer has been explicitly shared by the owner of the answer (shared_with: context) and
-    // if the user who is trying to read the answer is member of the same context as the author.
+    function researcherOfContext() {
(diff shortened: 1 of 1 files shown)


firestore.rules
```diff
@@ -25,8 +25,12 @@ service cloud.firestore {
         && request.auth.token.platform_id == resource.data.platform_id;
     }
 
-    // Check if answer has been explicitly shared by the owner of the answer (shared_with: context) and
-    // if the user who is trying to read the answer is member of the same context as the author.
```
pjanik, line 29: Wasn't that true anymore?

pjanik (approved):
Looks good 👍

firestore.rules
dougmartin: Good catch.  I created these rules on `report-service-dev` for development but for this PR I diffed the `report-service-pro` and `report-service-dev` rules to ensure there were no manually applied changes on pro and that comment was on pro and not dev but I did not copy over the comment.  I've added it back.

PR #211, opened 2024-03-31
feat: Add open response url in the details report using sourceKey fallback [PT-181870149]

dougmartin:
This adds an url to view the open response via the portal dashboard, mainly to hear the recorded audio answer.  Unlike newer question interactives open response only records the text as the answer and not the full interactive state.  Because of this there is no way to check if the open response has an audio answer.  Furthermore if the open response only has an audio answer there is no recorded text answer.

To add the url two things had to be changed with the details report:

1. Instead of joining learners to answers the order was switched and answers are joined to learners.  This allows learners that have only audio responses and no text response to still show in the report.
2. If the open response has no answer fall back to pulling the answerSourceKey parameter needed by the portal to first the answerSourceKey parameter of the interactive and if that isn't present the host of the interactive.

NOTE: this also fixes a bug in the managed_interactive/mw_interactive json value selector.

### DIFF
--- a/query-creator/create-query/steps/aws.js
+++ b/query-creator/create-query/steps/aws.js
@@ -140,46 +140,74 @@ const getColumnsForQuestion = (questionId, question, denormalizedResource, authD
 
   const columnPrefix = `res_${activityIndex}_${questionId}`;
 
-  const modelUrl = [`CONCAT(`,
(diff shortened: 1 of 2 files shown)


scytacki (approved):
looks good to me. :+1:

PR #212, opened 2024-04-24
Added start of Elixir-based report server [PT-187486745]

dougmartin:
This is mostly just the default output of running this project generator:

mix phx.new . --module ReportServer --app report_server --no-ecto --no-mailer

After running that the homepage and page title generator was customized and the favicon.ico was updated.


pjanik (approved):
Looks good 👍

PR #218, opened 2024-05-01
fix: Remove env from config and instead base it on portal url [PT-187509988]

dougmartin:
This mirrors what the existing report-service client app does.


pjanik (approved):
Looks good 👍

PR #219, opened 2024-05-01
feat: Added demo mode to server [PT-187514645]

dougmartin:
This uses faked responses from the portal and AWS to allow for faster development and test work.


pjanik (approved):
Looks good 👍

PR #224, opened 2024-05-15
feat: Add report-service api endpoint to return full answer document [PT-187486891]

dougmartin:
This is needed by the audio translation report post processing to determine if the answer has audio and where the path to the audio file.


pjanik (approved):
Looks good 👍

PR #225, opened 2024-05-15
feat: Added report_service module to server [PT-187486891]

dougmartin:
This will be used by the report post processing to get the answer from the report service api.


pjanik (approved):
Looks good 👍

PR #230, opened 2024-05-17
feat: Added copy url buttons [PT-187621671]

dougmartin:
This also switches all the buttons to use the small form factor.


pjanik (approved):
Looks good 👍

PR #238, opened 2024-05-20
Added /config endpoint to expose non-secret configuration options

dougmartin:
I needed to add this to figure out why the production server wasn't walking to the production token service.  You'll see the fix for that in TokenService#get_env.


pjanik (approved):
Looks good 👍

PR #239, opened 2024-05-21
build: Fixed production release issues

dougmartin:
- Added REPORT_SERVICE_URL environment variable (was defaulting to staging)
- Updated production user policy in readme to use production account for the transcription jobs
- Added report service url to config output


pjanik (approved):
Looks good 👍

PR #240, opened 2024-05-21
fix: Post processing step checkboxes not accepting initial click [PT-187486891]

dougmartin:
The form was not being updated within the query but rather set at the top level assigns.


pjanik (approved):
Looks good 👍

PR #241, opened 2024-05-21
fix: Change Url to URL in report-server UI [PT-187621671]

dougmartin:
This updates the button text to use URL instead of Url.


pjanik (commented):
Looks good 👍

pjanik (approved):

PR #242, opened 2024-05-21
fix: Unlinked job server from liveview process [PT-187621649]

dougmartin:
With this change the job server does not die when the LiveView that started it terminates.  Instead the LiveView registers itself and the server polls every minute and terminates itself when there are no attached clients and no running jobs.


pjanik (approved):
Looks good 👍

PR #244, opened 2024-05-22
fix: Copy prompt headers to new report columns [PT-187644206]

dougmartin:
This copies the prompt header from the open response question to all the new columns added in the post processing steps.


pjanik (approved):
Looks good 👍

PR #245, opened 2024-05-22
feat: Add support for hiding student names [PT-184362179]

dougmartin:
This change hides the student names in the details, usage and "wide" log report if the signed request params from the server contains a value for "hide_names".


pjanik (approved):
Looks good 👍

PR #248, opened 2024-05-23
feat: Remove log data username in wide learner logs [PT-187653588]

dougmartin:
This also adds missing tests for the log reports.


pjanik (approved):
Looks good 👍

PR #257, opened 2024-07-09
feat:  Add placeholder for glossary audio link post-processing step [PT-187885500]

dougmartin:
This is the start of the glossary audio link post processing step.  It defines the step and, for now, always adds the bare audio demo link to the csv.


pjanik (approved):
Looks good 👍

PR #258, opened 2024-07-15
feat: Add api endpoint to get plugin states [PT-187885524]

dougmartin:
This returns a (possibly empty) map of document ids to parsed plugin states for a remote endpoint under a source.


pjanik (approved):
Looks good 👍

PR #259, opened 2024-07-15
feat: Add glossary data column [PT-187885602]

dougmartin:
Updates glossary post processing job to just output a single column with the learners glossary data.

This is an interim step to get the raw data.  Later PRs will convert this raw data into two columns, one for the definitions and one for the link to the audio listening page.


pjanik (approved):
Looks good 👍

PR #266, opened 2024-10-30
feat: Add summary portal stats to homepage [PT-188475412]

dougmartin:
This adds a genserver that gathers summary stats from a list of portals and then periodically updates those stats.

The summary stats are shown on the homepage.

Here is a screenshot of the output:
image


bgoldowsky (approved):
We did a synch review, looks great.

PR #268, opened 2024-10-31
feat: Added start of CODAP plugin [PT-188507969]

dougmartin:


server/lib/report_server/portal_dbs.ex
```diff
@@ -0,0 +1,79 @@
...
+
+      GenServer.stop(pid)
+
+      result
+    else
+      error -> error
+    end
+  end
+
+  # converts columns: ["foo", "bar"], rows:[[1, 2], ...] to [%{"foo" => 1, "bar" => 2}, ...]
```
bgoldowsky, line 34:
```suggestion
  # converts columns: ["foo", "bar"], rows:[[1, 2], ...] to [%{:foo => 1, :bar => 2}, ...]
```

server/lib/report_server_web/live/codap_plugin_live/index.ex
```diff
@@ -0,0 +1,90 @@
...
+    "new_class_counts_by_year" => %Query{name: "New Class Counts By Year", sql: Query.count_sql("portal_clazzes", "classes", :year)},
+    "new_activity_counts_by_month" => %Query{name: "New Activity Counts By Month", sql: Query.count_sql("external_activities", "activities", :month)},
+    "new_activity_counts_by_year" => %Query{name: "New Activity Counts By Year", sql: Query.count_sql("external_activities", "activities", :year)},
+    "new_assignment_counts_by_month" => %Query{name: "New Assignment Counts By Month", sql: Query.count_sql("portal_offerings", "assignments", :month)},
+    "new_assignment_counts_by_year" => %Query{name: "New Assignment Counts By Year", sql: Query.count_sql("portal_offerings", "assignments", :year)},
+  }
+
+  @impl true
+  def mount(_params, _session, socket) do
+    socket = socket
```
bgoldowsky, line 33: I think `socket` here is what you were calling `state` in the home page live view.
Not sure why `socket` is used here?

bgoldowsky (approved):
This looks reasonable, and I saw a version of it working in your demo.
A couple of small questions inline.

server/lib/report_server_web/live/codap_plugin_live/index.ex
dougmartin: The socket is different from the state.  The socket is passed to use by the liveview router plug on each page load and is a struct that is analogous to the `conn` struct for regular requests that contains connection information.  

You can see all the elements of the socket struct here:

https://github.com/phoenixframework/phoenix_live_view/blob/main/lib/phoenix_live_view/socket.ex#L62-L73

within that struct there is an `assigns` map that is used in the same manner as the state in a genserver - it holds the current liveview state and anything in that map is available in the templates using the `@` prefix.

server/lib/report_server/portal_dbs.ex
dougmartin: Good catch, I'll update the comment.

PR #269, opened 2024-11-01
feat: Added env var to disable stats server

dougmartin:
The stats server currently won't work on staging due to security group issues so this lets us disable it there.


bgoldowsky (approved):
Looks good to me. Tested locally and don't see any connection attempts with the environment variable.

PR #270, opened 2024-11-01
build: Added npm install to Dockerfile

dougmartin:
Since the CODAP plugin uses npm to get the plugin api the release Dockerfile needs to have npm available


bgoldowsky (approved):
Looks good.
It would be helpful to add a line to the README "development setup" section as well.

dougmartin:
> It would be helpful to add a line to the README "development setup" section as well.

This change is baked in to the Dockerfile and only used for releases.

PR #271, opened 2024-11-01
fix: Disable ngss server for now

dougmartin:
The security ingress rules on production do not currently allow the report server to connect to the ngss db so it is being temporarily removed from the list of servers.

NOTE: the jump from `pre-1` to `pre-3` was because I had updated staging and production to `pre-2` and then needed to make this fix and so I needed a new docker image tag to redeploy it.


bgoldowsky (approved):
LGTM 👍

bgoldowsky:
Alternatively, servers could reasonably be set as an environment variable.

PR #275, opened 2024-11-07
feat: Add MySQL database and user table [PT-188527978]

dougmartin:
This adds Ecto with the MySQL driver to the code along with an initial user table that is used in the oauth code to track the user.

The docker-compose.yml file has been changed to only host a MySQL db.

This commit also adds a user test and fixes all the existing tests.


bgoldowsky (approved):
Synch reviewed.

PR #276, opened 2024-11-07
Fix bug where dates don't sort properly

bgoldowsky:
For some reason Elixir Date objects don't compare properly with <, > etc.
So provide a special case to sort them.

### DIFF
--- a/server/lib/report_server/portal_dbs.ex
+++ b/server/lib/report_server/portal_dbs.ex
@@ -42,14 +42,9 @@ defmodule ReportServer.PortalDbs do
     end)
   end
 
-  defp value_sorter(v1, v2) do
-    # Dates do not sort properly with normal <= operator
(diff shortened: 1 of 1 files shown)


server/lib/report_server/portal_dbs.ex
```diff
@@ -42,10 +42,20 @@ defmodule ReportServer.PortalDbs do
     end)
   end
 
+  defp value_sorter(v1, v2) do
```
dougmartin, line 45: You could replace this with something like this to pattern match on the stuct type of the params:

```
  defp value_sorter(v1 = %Date{}, v2 = %Date{}), do: Date.compare(v1, v2) != :gt
  defp value_sorter(v1, v2), do: v1 <= v2
```

dougmartin (approved):
Looks good :+1:

PR #278, opened 2024-11-13
feat: Add start of report taxonomy [PT-188507979]

dougmartin:


bgoldowsky (approved):
This looks good.  I was able to test it and see the hierarchy in action (but we should add the requirement to create the dev database to the README).

PR #283, opened 2024-11-14
feat: Added __using__ macro to reports

dougmartin:
This will cut down on the boilerplate in the reports.


bgoldowsky (approved):
Looks good!

PR #284, opened 2024-11-14
feat: Disable report tree cache in dev mode

dougmartin:
This prevents the run function of each report from being cached in memory which would require a restart of the server on any change to that function.


bgoldowsky (approved):
Makes sense.

PR #287, opened 2024-11-15
feat: Added filter queries [PT-188527665]

dougmartin:
Added dynamic filter queries and updated report tree with TBD report placeholders.


bgoldowsky (approved):
Approving; there are a few known issues that will be taken care of in later PRs.

PR #288, opened 2024-11-15
Query refinement

bgoldowsky:
Updates the three report queries.

Minor changes to the teacher report, but large rewrites to the resource summary and detail.

Properly uses the tables to look at offerings, rather than leaning on the report_learners table for everything.

One remining issue is determining grade level.  Not sure if the query is wrong or if the canned samples simply do not have grade levels specified.

Updates names and descriptions of the reports to what is in the design doc.


dougmartin (approved):
Looks good :+1:

PR #289, opened 2024-11-15
Header tweaks

bgoldowsky:
Display real name rather than username (which is arbitrary for google logins).
Show if user is admin.
Underline links.


dougmartin (approved):
Looks good :+1:

PR #290, opened 2024-11-15
fix: Update resource metrics summary report to use left join

dougmartin:
This ensures resources that have not been assigned are still shown.


bgoldowsky (approved):
Looks good.

PR #293, opened 2024-11-20
First draft of date filtering

bgoldowsky:
This adds date pickers to the UI, and filters the three existing queries using those dates to bound `last_run`.
Probably needs some tweaks but ready for a review to see if it's on the right track.


dougmartin (changes requested):
Looks like a good start!

server/lib/report_server_web/live/new_report_live/form.ex
```diff
@@ -45,6 +45,8 @@ defmodule ReportServerWeb.NewReportLive.Form do
     |> assign(:filter_type_options, [@filter_type_options])
     |> assign(:filter_options, [[]])
 
+    {:ok, socket}
```
dougmartin, line 48: You can remove this line - the `handle_params` returns `{:noreply, socket}` on line 50.

dougmartin (approved):
Looks good :+1:

PR #295, opened 2024-11-22
Permission form filtering

bgoldowsky:
PT-188589623

Adds 'Permission forms' to the set of filters supported.  Includes appropriate queries for permission forms to be limited by other filters and for other filters to be limited by permission forms.

Updates some of the other quries to avoid joining the same field twice where not needed.

Adds a 'filters_included' field to reports so not all existing filters need to be included on each report.  The 3 existing reports do not use Permissions forms filtering.

Improve error handling in `get_sql`.  Now returns a tuple with `:ok` or `:error`.

Adds some (not exhaustive) tests for `ReportQuery` and `ReportFilterQuery`.

**Skips** 7 tests that were failing.


dougmartin (approved):
Looks great :+1:

PR #296, opened 2024-11-25
Filter UI improvements

bgoldowsky:
Update/set the live_select list of options when the filter type is changed.

Put date filters on one line.


dougmartin (approved):
Looks good :+1:

PR #297, opened 2024-11-26
Limit how many rows of report on shown in live view.

bgoldowsky:
PT-188589069: only show the first 100 rows of queries.

No longer stores the full results of queries in memory after they are displayed.
This means query has to be re-run for sort changes and downloads.

We want the database server to do the query with a `limit`, which means it has to do the sorting as well so that we get the correct top rows.  So this includes a refactor to construct the appropriate `order by` clauses for any choice of sorting.

A separate query is contructed and executed (once, since sorts/downloads shouldn't affect it) to determine the total number of rows.

Also refactored the `ReportQuery` struct to have a list of cols rather than a single `select` string. This turned out not to be necessary but seems like it might be useful in the future so I kept it in.

Testing note: you can set `@row_limit 10` in `show.ex` to have the limit take effect more often.

### DIFF
--- a/server/lib/report_server/reports/report_query.ex
+++ b/server/lib/report_server/reports/report_query.ex
@@ -1,9 +1,9 @@
 defmodule ReportServer.Reports.ReportQuery do
   alias ReportServer.Reports.ReportQuery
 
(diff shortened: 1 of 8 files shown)


dougmartin (approved):
Looks great! :+1:

PR #300, opened 2024-12-03
feat: Add Athena reports [PT-188587537]

dougmartin:
Adds support for generating and tracking athena queries with the student actions as the first report implemented.

### DIFF
--- a/server/lib/report_server/reports/report_query.ex
+++ b/server/lib/report_server/reports/report_query.ex
@@ -16,14 +16,14 @@ defmodule ReportServer.Reports.ReportQuery do
     end
     limit_sql = if limit != nil, do: "LIMIT #{limit}", else: ""
 
-    "SELECT #{select_sql} FROM #{from} #{join_sql} WHERE #{where_sql} #{group_by_sql} #{order_by_sql} #{limit_sql}"
+    {:ok, "SELECT #{select_sql} FROM #{from} #{join_sql} WHERE #{where_sql} #{group_by_sql} #{order_by_sql} #{limit_sql}"}
   end
 
   def get_count_sql(%ReportQuery{from: from, join: join, where: where, group_by: group_by}) do
     query_without_cols_or_order = %ReportQuery{cols: [{"1", "row"}], from: from, join: join, where: where, group_by: group_by}
-    subquery = get_sql(query_without_cols_or_order)
+    {:ok, subquery} = get_sql(query_without_cols_or_order)
 
-    "SELECT COUNT(*) AS count FROM (#{subquery}) AS subquery"
+    {:ok, "SELECT COUNT(*) AS count FROM (#{subquery}) AS subquery"}
   end
 
   def update_query(report_query = %ReportQuery{}, opts) do
@@ -51,4 +51,25 @@ defmodule ReportServer.Reports.ReportQuery do
     %{report_query | order_by: new_order_by}
   end
 
+  def get_log_cols(opts) do
+    hide_names = Keyword.get(opts, :hide_names, false)
+    remove_username = Keyword.get(opts, :remove_username, false)
+
+    ["id", "session", "username", "application", "activity", "event", "event_value", "time", "parameters", "extras", "run_remote_endpoint", "timestamp"]
+      |> Enum.filter(&(!(&1 == "username" && remove_username)))
+      |> Enum.map(&(maybe_hash_username(&1 == "username" && hide_names, &1)))
+  end
+
+  def get_log_db_name() do
+    Application.get_env(:report_server, :athena) |> Keyword.get(:log_db_name, "log_ingester_production")
+  end
+
+  defp maybe_hash_username(hash, col) do
+    if hash do
+      hide_username_hash_salt = Application.get_env(:report_server, :athena) |> Keyword.get(:hide_username_hash_salt, "no-hide-username-salt-provided!!!");
+      {"TO_HEX(SHA1(CAST(('#{hide_username_hash_salt}' || \"log\".#{col}) AS VARBINARY)))", col}
+    else
+      {"\"log\".\"#{col}\"", col}
+    end
+  end
 end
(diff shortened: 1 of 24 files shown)


server/lib/report_server/reports/report_query.ex
```diff
@@ -51,4 +51,25 @@ defmodule ReportServer.Reports.ReportQuery do
...
+  end
+
+  defp maybe_hash_username(hash, col) do
+    if hash do
+      hide_username_hash_salt = Application.get_env(:report_server, :athena) |> Keyword.get(:hide_username_hash_salt, "no-hide-username-salt-provided!!!");
+      {"TO_HEX(SHA1(CAST(('#{hide_username_hash_salt}' || \"log\".#{col}) AS VARBINARY)))", col}
+    else
+      {"\"log\".\"#{col}\"", col}
+    end
+  end
```
bgoldowsky, line 74: I think the athena data that gets appended has multiple IDs for the user already -- `user_id` and `student_id`.  I suggested in [this document](https://docs.google.com/document/d/16W3Mdq8x7gXs-XilEzic0wIwC4je8tuAyTqiqJ5mrFs/edit?tab=t.0) that it would be nice to standardize on one. Is there a reason that we want username in particular (hashed or not)?

server/lib/report_server/reports/student_actions.ex
```diff
@@ -0,0 +1,88 @@
...
+  alias ReportServer.PortalDbs
+
+  def get_query(report_filter = %ReportFilter{}, user = %User{}) do
+    run_remote_endpoints = get_run_remote_endpoints(report_filter, user)
+      |> Enum.map(&("'#{&1}'"))
+      |> Enum.join(", ")
+
+    # FIXME: when we add non-admin access update this code to only allow admins to not hide names
+    hide_names = report_filter.hide_names
+    remove_username = false
```
bgoldowsky, line 13: What is the difference between these two options?

server/lib/report_server/reports/tree.ex
```diff
@@ -124,10 +125,12 @@ defmodule ReportServer.Reports.Tree do
         })
       ]},
       %ReportGroup{slug: "student-reports", title: "Student Reports", subtitle: "Reports about students", tbd: true, children: [
-        TBDReport.new(%Report{
+        StudentActions.new(%Report{
```
bgoldowsky, line 128: Should we make the "student-reports" Group no longer `:tbd` now?

server/lib/report_server_web/live/new_report_live/form.html.heex
```diff
@@ -73,16 +73,28 @@
...
-    </div>
-    <div class="mt-4">
-      <.button :if={!blank?(@form.params["filter1"])} type="button" phx-click="submit_form"><.icon name="hero-bolt" /> Run Report</.button>
-    </div>
+      </div>
+
+      <div :if={@form_options.enable_hide_names}>
+        <div class="flex gap-2 items-center">
+          <.input type="checkbox" id="hide_names" field={@form["hide_names"]} />
+          <label for="hide_names" class="whitespace-nowrap">Hide names</label>
```
bgoldowsky, line 90: For improved privacy, I'd suggest we hide names by default for everyone, and admins have to check a 'Show names' option to see them.

server/lib/report_server_web/live/report_run_live/show.ex
```diff
@@ -140,4 +172,64 @@ defmodule ReportServerWeb.ReportRunLive.Show do
...
+          report_run
+      end
+    else
+      report_run
+    end
+  end
+
+  defp maybe_poll_query_state(report_run = %ReportRun{}, live_view_pid) do
+    if poll_query_state?(report_run) do
+      Process.send_after(live_view_pid, :poll_query_state, 1000)
```
bgoldowsky, line 192: Is polling every second a lot?  Should we do some sort of backoff for long queries?

server/lib/report_server_web/live/report_run_live/show.ex
```diff
@@ -97,18 +109,22 @@ defmodule ReportServerWeb.ReportRunLive.Show do
...
+         {:ok, athena_query_id, athena_query_state} <- AthenaDB.query(sql, report_run_id, report_run.user),
+         {:ok, _report_run} <- Reports.update_report_run(report_run, %{athena_query_id: athena_query_id, athena_query_state: athena_query_state}) do
+
+      send(live_view_pid, :poll_query_state)
+
+      # return nil - the template will use the report_run to provide status updates
+      {:ok, %{report_results: nil}}
 
     else
       {:error, error} ->
```
bgoldowsky, line 130: As noted in Slack, athena seems to return `:exit` rather than `:error` for at least some kinds of problems.

bgoldowsky (changes requested):
Mostly just questions. Marked as 'request changes' mostly since I can't test it & the errors I get are not handled smoothly.

server/lib/report_server/reports/report_query.ex
dougmartin: I think (at least for now to ease QA) we should have the reports generate the same output so they can be tested against the current reports from the portal to ensure they return the same info.  I think there is going to probably be a pass over the reports near the end of the work to clean them up.

server/lib/report_server/reports/student_actions.ex
dougmartin: That is in there for future work - I ported the code from Javascript and it is used in the joined learner reports.

server/lib/report_server/reports/tree.ex
dougmartin: Yes - I'll change that in my next PR.

server/lib/report_server_web/live/new_report_live/form.html.heex
dougmartin: Probably a good idea - we an update that in a future PR if Kiley agrees.

server/lib/report_server_web/live/report_run_live/show.ex
dougmartin: The 1 second pooling is what we are currently doing and it works fine.  The problem with a backoff is folks will be waiting for these to complete the if we backoff they will be waiting way more than they need to wait.

PR #302, opened 2024-12-04
doc: Add user policy docs

dougmartin:
Adds the user policy docs so they are captured somewhere as they are not present in any CloudFormation template.


bgoldowsky (approved):
I don't have the knowledge to review these or say if this is the right place to store them, but it's clearly better than nothing since storing them somewhere is a good idea.

PR #305, opened 2024-12-05
dev: Add server logs to the browser console

dougmartin:
This uses a new feature in phoenix_live_reload to inject the server logs into the client console for easier debugging.  This only is enabled in dev.


bgoldowsky (approved):
Seems helpful!

PR #306, opened 2024-12-05
dev: Reorganize reports into sub-folders

dougmartin:
This moves the portal and athena reports into subfolders for better organization and appends "Report" to the report modules.


bgoldowsky (approved):
Nice.

PR #307, opened 2024-12-10
Change resource detail report grades column to class grades

bgoldowsky:
Changes the query to show the grades of the classes to which the resource has been assigned, rather than the grades associated with the reource itself.


dougmartin (approved):
Looks good :+1:

PR #309, opened 2024-12-16
Fix more ux issues

bgoldowsky:
PT-188527665

Fix a few UI problems.
- If a filter exists, but no values are chosen, return no results, instead of a potentially very long list.
- Reset selection menu when filter type is changed.
- Include query string in placeholder text, for clarity
- Allow searching cohort with 1-2 characters, since all values are shown with no characters anyway
- Changing filter type removes list of options of previous type


dougmartin (approved):
Looks good :+1:

PR #311, opened 2024-12-17
fix: Reports broken when filter_list is nil

dougmartin:
Also fixed a couple of missed compiler warnings about unused variables and removed a debug IO.inspect that was left in the code.


bgoldowsky (approved):
LGTM 👍

PR #313, opened 2024-12-18
feat: Add report preview note to Athena reports [PT-188700907]

dougmartin:
This adds the following note to the Athena reports once they succeed:

"Note: Report preview and JSON download are not supported for this report type."

image


bgoldowsky (approved):
Looks good. ✅

PR #314, opened 2024-12-18
188613508 query runs table

bgoldowsky:
Use the `portal_runs` table for filtering reports by date, and add "Number of runs" and "First run" to Teacher Status and Detailed Metrics.

PT-188613525
PT-188613508


dougmartin (approved):
Looks good :+1:

PR #316, opened 2024-12-19
Add exclude checkbox & support

bgoldowsky:
PT-188645467

Add a checkbox and update queries to exclude teachers with @concord.org emails.


dougmartin (approved):
Looks great :+1:

PR #317, opened 2024-12-20
feat: Added details and usage reports [PT-188527607] [PT-188527655]

dougmartin:
Added the details and usage reports, converted from the JavsScript version to Elixir.

This also moves the `portal_report.ex` and `report_service.ex` files to the `report_server` directory from the `report_server_web` directory.


bgoldowsky (approved):
Sync reviewed

PR #318, opened 2025-01-02
Fix number of runs in reports

bgoldowsky:
Addresses QA comment in PT-188613525

Count of number of runs in details report was wrong due to lack of `distinct` in query.
Teacher status report appeared to be correct but adding `distinct` on runs there as well in case untested cases or future updates cause retrieval of multiple rows with the same run id.


dougmartin (approved):
Looks good :+1:

PR #319, opened 2025-01-02
fix: Added check for no learners [PT-188527607]

dougmartin:
This fix causes the query to exit early with a "No learner data found" error when there are no learners found matching a query.


bgoldowsky (approved):
LGTM! 👍

PR #321, opened 2025-01-15
fix: Resource ordering in shared queries [PT-188527607]

dougmartin:
Converts learner and resource data from a map to a list and adds a question_order field to the denormalized resource.

The original JavaScript used a hash where the key ordering is maintained in insertion order.  In Elixir the key order is alphabetic so lists must be used to maintain ordering of both resources and questions within denormalized resources.

This also fixes a deprecation warning in the Floki library when extracting the text from the student answer.


bgoldowsky (approved):
Can't say I understand every detail of this change, but overall looks good!

PR #322, opened 2025-01-16
fix: Crash on large requests [PT-188748565]

dougmartin:
This fixes a bug in the code that generated the question columns in the shared query.  The inner loop of the reducer was mapping its result instead of reducing on it.  This caused enormously large nested lists to be created when there were a large number of columns.  These nested lists caused geometric slowdown of each reduce pass as the lists increased in size.

The fix for this was to reduce instead of mapping in the inner reducer - this causes the list to only increase in side proportional to the number of columns.

I also added a flag to skip retries of the resource if the Firebase function returns 500, which it does when the function does not provide a source parameter that is correct.


bgoldowsky (approved):
Looks good!

PR #323, opened 2025-01-17
fix: Accidentally removes List.reverse call [PT-188527607]

dougmartin:
This line was accidentally removed while debugging the column ordering issue.


bgoldowsky (approved):
👍

PR #324, opened 2025-01-17
feat: Add check for max sql length [PT-188748565]

dougmartin:
Athena has a max sql query size of 256KB which should be checked before the query is submitted to Athena.


bgoldowsky (approved):
👀 👍

PR #325, opened 2025-01-21
feat: Added post processing to new report [PT-188613581]

dougmartin:
This integrates the existing post-processing facilities with the new reports.  The jobs are still saved on S3 instead of using a new table so that the existing post-processing from the portal launched reports work without updates.


bgoldowsky (approved):
This looks good to me.

PR #326, opened 2025-01-22
feat: Added CLUE link to work post processing step [PT-188520218]

dougmartin:
This required post processing of the report to extract the learners.  The post processing is setup to allow for other future post processing steps if needed.


bgoldowsky (approved):
Thank you for the documentation comments. This looks good.

PR #327, opened 2025-01-22
fix: Crash when no filters and "Exclude Concord Consortium user accounts" is checked [PT-188805157]

dougmartin:


bgoldowsky (approved):
Looks good!

PR #328, opened 2025-01-22
fix: Move /reports to /old-reports [PT-188779638]

dougmartin:
This is the first step in a three step process to remove the old reports and replace the /reports url with the new reports.

Once this is live the existing query-creator Firebase function can be updated to redirect to /old-reports.  Once that is live then a follow on PR will rename /new-reports to /reports.

Then, the portals can then be updated to remove the report links from the admin and add the report link in the sidebar to point to /reports.

Finally, a follow on PR will remove all the /old-reports code that is no longer needed.


bgoldowsky (approved):
Changes look fine 👍

dougmartin:
@bgoldowsky Thanks - I'm going to set this as a draft so it isn't accidentally merged.  I want to wait to merge until we have cleared QA.

PR #329, opened 2025-01-23
fix: Add targetUserId to CLUE url [PT-188520218]

dougmartin:
This is needed so the portal can verify the researcher has access to the log event.


bgoldowsky (approved):
LGTM 👍

PR #330, opened 2025-01-24
feat: Move /new-reports to /reports [PT-188779638]

dougmartin:
This updates all /new-reports links to /reports and adds a plug to automatically redirect /new-report urls to /reports so test links don't break.


bgoldowsky (approved):
LGTM 👍

PR #334, opened 2025-01-24
Fix column name in query

bgoldowsky:
PT-188818747

Fixes column name which was `aci` in JOIN clause but `aci_cohort` in ON clause.


dougmartin (approved):
Looks good :+1:

PR #335, opened 2025-01-25
build: Update report link in query-creator sam config [PT-188779638]

dougmartin:
Updated the researcher reports url that is redirected to after the query is created to the /old-reports link.


bgoldowsky (approved):
Looks good!

PR #337, opened 2025-01-27
fix: Add user check to homepage [PT-188779638]

dougmartin:
This fixes the homepage when there is no logged in user.


bgoldowsky (approved):
👍

PR #342, opened 2025-02-03
feat: Added student feedback metadata api endpoint [PT-188732128]

dougmartin:
This returns a map of offering ids to an object that, for now, only contains the last updated at time in seconds.


emcelroy (approved):
Looks good 👍

PR #347, opened 2025-02-21
Add primary_user_id to reports

bgoldowsky:
This adds a `primary_user_id` column to the **Student Answers**, **Student Actions**, and **Student Actions with Metadata** reports.

**Student Actions** did not previously have any columns other than what is in the log data itself. For that report this PR adds the `user_id` column as well as `primary_user_id`. In order to do this the report is refactored to work like *Student Actions with Metadata** but with just the minimal 2 columns being joined on from portal data, rather than the longer list of metadata columns.


dougmartin (approved):
Looks good :+1:

PR #349, opened 2025-03-10
Add CLUE text tile content to answers report

bgoldowsky:
REPORT-2


server/lib/report_server/clue.ex
```diff
@@ -0,0 +1,207 @@
...
+        answers_df = Explorer.DataFrame.new(answerlist)
+        Explorer.DataFrame.to_parquet(answers_df, path)
+      else
+        _ -> IO.inspect({:error, "Failed to construct parquet file path"})
+      end
+    end)
+    if (Enum.all?(write_attempts, fn result -> result == :ok end)) do
+      {:ok, result}
+    else
+      IO.inspect({:error, "Failed to write parquet files"})
```
dougmartin, line 190: I think you should return `{:error, "Failed to write parquet files"}` here.

server/lib/report_server/clue.ex
```diff
@@ -0,0 +1,207 @@
...
+    bucket = System.get_env("ATHENA_REPORT_BUCKET")
+    escaped_url = ReportUtils.escape_url_for_filename(url)
+    # The 'username' field will be something like "user_id@portal_site".
+    [user_id, portal_site] = String.split(username, "@")
+    platform_id = ReportUtils.escape_url_for_filename("https://#{portal_site}")
+    path = "s3://#{bucket}/partitioned-answers/#{escaped_url}/#{platform_id}/#{resource_link_id}/#{user_id}.parquet";
+    FSS.S3.parse(path, config: get_s3_config())
+  end
+
+  defp get_s3_config() do
```
dougmartin, line 84: You could also use the `runtime.exs` configuration.  In the future we should clean that up (by removing the dead configs from the old athena query handler) and use that here.

server/lib/report_server/clue.ex
```diff
@@ -0,0 +1,207 @@
...
+        answers: updated_answers
+      }
+    end)
+
+    ## Loop over answers and write a parquet file for each username
+    write_attempts = Enum.map(result.answers, fn {username, answerlist} ->
+      resource_link_id = answerlist |> List.first() |> Map.get(:resource_link_id)
+      with {:ok, path} <- get_parquet_file_path(url, username, resource_link_id) do
+        answers_df = Explorer.DataFrame.new(answerlist)
+        Explorer.DataFrame.to_parquet(answers_df, path)
```
dougmartin, line 182: The path here is an `s3://` path right?  Writing to the filesystem directly isn't going to work on production as the ECS docker containers can't write files locally.

dougmartin (changes requested):
Looks great! Just one change to return the `{:error, ...}` tuple.

dougmartin (approved):
Looks good :+1:

PR #350, opened 2025-03-10
Fix error in count SQL

bgoldowsky:
After MySQL upgrade on staging, count of rows at the top of portal report pages is showing an error.

Seems to be because "row" is a reserved word in MySQL, which may be more strictly enforced in v8.


dougmartin (approved):
Looks good :+1:

PR #358, opened 2025-03-27
fix: Exclude CC users learner data query [REPORT-29]

dougmartin:
Previously the query was using a "NOT LIKE '%@concord.org'" filter which both caused a full table scan of the users table (which is very large in production).  This caused the query to timeout.

The change is to instead query the database separately if the exclude CC users setting is set to get a list of all teacher ids in schools with "Concord Consortium" in their name and then join against the portal teachers table and exclude those teachers from the result.  This was the original behavior of the exclude CC users checkbox in the old reports.

### DIFF
--- a/server/lib/report_server/reports/athena/learner_data.ex
+++ b/server/lib/report_server/reports/athena/learner_data.ex
@@ -91,15 +91,20 @@ defmodule ReportServer.Reports.Athena.LearnerData do
       {join, where}
     end
 
-    {join, where} = if exclude_internal do
+    internal_teacher_ids = if exclude_internal do
+      get_internal_teacher_ids(user.portal_server)
+    else
+      []
+    end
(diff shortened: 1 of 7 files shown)


bgoldowsky (commented):
I think you should update the `exclude_internal_accounts` method on the filter side to match this logic, otherwise there could be inconsistencies between the teachers that show up in the query interface and the ones that show up in the report.

dougmartin:
@bgoldowsky Good point - I'll update to use it there.

dougmartin:
@bgoldowsky Fixed in https://github.com/concord-consortium/report-service/pull/358/commits/7eef4b5e19a9bbeb9feda592d90b2094c110e174

bgoldowsky (approved):
Looks good!

PR #360, opened 2025-03-28
fix: Remove demo mode [REPORT-15]

dougmartin:
Remove demo mode from server along with the entire concept of modes.

Also fixed compiler warnings including swapping parameter order of exclude_internal_accounts so that the default value could be used in a function head to remove the warning.


bgoldowsky (approved):
This seems like a good simplification. 👍

PR #362, opened 2025-03-31
build: Add GitHub action to run server mix tests

dougmartin:


.github/workflows/report-server.yml
```diff
@@ -0,0 +1,60 @@
+name: Report Server Tests
+
+on:
+  push:
+    paths:
+      - 'server/**'
+      - '.github/workflows/**'
```
dougmartin, line 7: I'll remove this in a follow on PR - it is needed so the tests run for this PR that setup the tests :)

dougmartin:
@bgoldowsky I retried reversing the ordering of the PRs to make the new test PR chained after this test but there are warnings in `master` that were fixed in that PR that prevent this PR from running the ci test.

bgoldowsky (approved):
Looks good

PR #363, opened 2025-04-02
feat: Add merge to primary user postprocessing step [REPORT-15]

dougmartin:
This adds a preprocessing step to the details report that gathers the primary user info in a preprocessing step and then uses that info to output the final csv.


server/lib/report_server/post_processing/job.ex
```diff
@@ -43,16 +49,43 @@ defmodule ReportServer.PostProcessing.Job do
               # the accumulator is nil on the first line
```
bgoldowsky, line 49: outdated comment?

server/lib/report_server/post_processing/job.ex
```diff
@@ -43,16 +49,43 @@ defmodule ReportServer.PostProcessing.Job do
...
               header_map = Helpers.get_header_map(row)
               rre_index = header_map["run_remote_endpoint"]
-              # add other index lookups here for future actions
-              {[new_preprocessed_row()], %{rre_index: rre_index}}
+              user_id_index = header_map["user_id"]
+              primary_user_id_index = header_map["primary_user_id"]
+              resource_header_map = Enum.reduce(header_map, %{}, fn {k,v}, acc ->
+                if String.starts_with?(k, "res_") do
```
bgoldowsky, line 55: This string matching feels fragile. At least, I guess we should document somewhere that reports can't have any other column that fits this pattern.

server/lib/report_server/post_processing/job.ex
```diff
@@ -24,9 +24,15 @@ defmodule ReportServer.PostProcessing.Job do
...
+  # NOTE: if you add another preprocessing action here, this should be refactored to
+  # put the preprocessing in each step instead of it all being here - the learner
+  # preprocessing should be moved to a helper module that could be imported into the step
+  # so that it could be used for multiple steps
   defp preprocess_rows(job, query_result, overrides) do
     preprocessed = %{
-      learners: %{}
+      learners: %{},
+      merged_user_ids: %{},
+      user_resources: %{},
```
bgoldowsky, line 35: Could you document the structures that are expected to end up in each of these maps? The code is difficult to follow to understand what is happening in these preprocessing steps, and what the expected output might be.

bgoldowsky (approved):
Approving so that you can move on to project review and testing.
I wasn't really able to determine if the code will work since it's pretty complex and undocumented. However, I think any errors will only affect the new postprocessing step so it should be safe to deploy and test on the server.

server/lib/report_server/post_processing/job.ex
dougmartin: Not outdated but definitely vague - I've updated it in https://github.com/concord-consortium/report-service/pull/363/commits/8a44ee9d41734780144fd034dea3eee1c21c4e29.

server/lib/report_server/post_processing/job.ex
dougmartin: I've updated it to use a regex to include a digit after the `res_` to make it a little more specific.  I agree it is a little brittle but we control both ends so I think the risk is minimal.

PR #366, opened 2025-05-13
feat: Add a user id prefix to merged user answers [REPORT-15]

dougmartin:
This changes the merged primary user_id answers from a list of answers joined by a comma to a list of answers with a user_id prefix per answer joined by newlines.  This will allow researchers to see which primar/non-primary user answered each question.

If there are no separate answers per user there is no prefix added.

If the primary user is not found as a user in the set of rows the first user found that has that primary user_id is used.


bgoldowsky (approved):
LGTM! 👍

PR #368, opened 2025-07-31
feat: Added script to repartition log data and updated code to use it  [REPORT-42]

dougmartin:
- Added script to repartition by app, year, month and run_remote_endpoint
- Updated report server code to use new partitioned table
- Add AWS Batch job to run the partitioner script
- Updated version to 1.5.0


bgoldowsky (approved):
Looks good!

PR #396, opened 2026-07-15
cc-data: bulk-read answers + interactive history + attachments (paged, resumable) [REPORT-76]

dougmartin:
## Summary

STORY 3 of the cc-data researcher toolchain (REPORT-76). Adds authenticated, paged, resumable bulk-read endpoints so a researcher can download, for a report run they own, every student's raw saved answers, the full interactive-state history (the moment-by-moment trajectory), and the S3 attachments those answers reference — the detail the existing denormalized CSV report can't express. Every access is authorization-gated and audit-logged.

Implements the full [closed spec](https://github.com/concord-consortium/report-service/blob/REPORT-76-add-bulk-read-api/specs/REPORT-76-cc-data-bulk-read-answers-history.md).

## What's in it

**Elixir API**
- `GET /api/v1/reports/:id/answers` and `GET /api/v1/reports/:id/history` — ownership-gated, derive-once authorized endpoint snapshot (permission short-circuit, `allow_empty`, per-learner `SourceKey` derivation), proxy to Node, reassemble an opaque composite `page_token`.
- `POST /api/v1/reports/:id/attachments` — batch presigned-GET URLs; durable auth, server-cred presign, IDOR guard on the doc's `remote_endpoint`, per-item partial success.
- `export_scratch` table + `Exports` context: two-step 404-vs-410 read-time lookup, absolute sliding TTL, fail-open tuple cache, `EctoJsonArray` MyXQL round-trip, and a boot/periodic sweep GenServer.
- Fail-closed audit (`export_scoped` intent + per-page `bulk_read` access rows, `attachment_urls_issued`), correlated by `export_id`; admin audit-log page gains `export_id` + `remote_endpoint` filters with full a11y.

**Node / Firebase**
- `POST /bulk_read` walker for answers (`__name__`) and history (`created_at, __name__` post-filtered to `remote_endpoint`) under four caps — `limit` / `endpoint_limit` / `read_limit` / an ~8 MB response-byte budget — with a lookahead-proven `endpoint_exhausted` and mid-batch item/byte-cap stops (no gap/dup).
- `POST /fetch_attachment_meta` — authoritative attachment metadata, both behind a header-only bearer guard; shared `api` function timeout raised 60 → 300 s.
- First emulator-backed test harness (fail-closed setup, faithful seed helpers, `--runInBand`).

## Testing

- Elixir: **374 tests, 0 failures** (7 skipped — portal-DB-dependent)
- Node pure: **163 tests, 0 failures**
- Node emulator (real Firestore emulator): **18 tests, 0 failures**

Two independent adversarial reviews (Elixir + Node) found no correctness/security bugs; the one latent-crash note surfaced is fixed and tested.

## Deploy prerequisites (see the closed spec's Technical Notes)

- Create the `interactive_state_histories (platform_id, platform_user_id, resource_link_id, created_at)` composite index in the target project — the pre-existing `context_id`-led indexes cannot serve this query (missing → `FAILED_PRECONDITION`).
- The shared `api` `timeoutSeconds` bump (60 → 300) lifts the ceiling for **every** co-located route — call out in deploy review; sanity-check the ALB/proxy idle timeout against ~300 s.
- Two new migrations run on deploy (`export_scratch`, `data_access_log.export_id`).
- Run the env-gated real-data source-fidelity check once before exercising `/answers` + `/history` for real.

## Note for reviewers

`firebase-tools` was added to `functions/package.json` devDependencies but `functions/package-lock.json` was intentionally **not** regenerated — npm wanted to migrate the whole lockfile v2 → v3 (a large unrelated churn), and there is no `npm ci` job for `functions/` in CI. If a functions `npm ci` job is added later, run `npm install` in `functions/` to sync the lock.


server/lib/report_server_web/api/v1/bulk_export_controller.ex
```diff
@@ -0,0 +1,176 @@
...
+        case AuditLog.create_entry(access_attrs) do
+          {:ok, _entry} -> json(conn, %{items: items, next_page_token: next_token})
+          {:error, _reason} -> ErrorHelpers.server_error(conn)
+        end
+
+      {:error, _reason} ->
+        # Node read failure (or a collapsed Node 4xx from a hand-crafted cursor Elixir already pre-validates):
+        # no audit row, no cursor advance; the CLI retries the same token idempotently.
+        ErrorHelpers.server_error(conn)
+    end
```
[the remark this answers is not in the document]
dougmartin: Fixed in a3ed871 — added an `{:ok, unexpected}` clause that logs and returns a controlled `SERVER_ERROR` (500) instead of letting a `CaseClauseError` crash the request. Covered by a new controller test.

server/lib/report_server_web/api/v1/attachment_controller.ex
```diff
@@ -0,0 +1,71 @@
...
+          |> json(AttachmentJSON.results(results, Aws.presign_ttl_seconds()))
+
+        {:error, _} ->
+          ErrorHelpers.server_error(conn)
+      end
+    else
+      {:error, :bad_request, msg} -> ErrorHelpers.bad_request(conn, msg)
+      {:error, :not_found} -> ErrorHelpers.not_found(conn)
+      {:error, _} -> ErrorHelpers.server_error(conn)
+    end
```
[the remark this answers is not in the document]
dougmartin: Fixed in a3ed871 — added an `{:ok, _unexpected}` else clause mapping an unexpected Node body (missing `"results"`) to a controlled `SERVER_ERROR` (500) rather than raising `WithClauseError`. Covered by a new controller test.

functions/src/api/bulk-read.ts
```diff
@@ -0,0 +1,274 @@
...
+  let hitItemCap = false;   // broke mid-batch on the item cap
+  let hitByteCap = false;   // broke mid-batch on the byte cap
+
+  while (true) {
+    if (items.length >= remItems || reads >= remReads) { exhausted = false; break; }
+    const batch = Math.max(1, Math.min(HISTORY_BATCH, remReads - reads));
+    let q = metaBase();
+    if (cursor) { q = q.startAfter(reconstructTimestamp(cursor), cursor.docId); }
+    const metaSnap = await q.limit(batch).get();
+    reads += metaSnap.size;
```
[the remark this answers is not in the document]
dougmartin: This overshoot is intentional and bounded — leaving as-is. The metadata query is deliberately the only thing sized against the remaining budget, so a page can exceed `read_limit` by at most one `HISTORY_BATCH` (~300, ~6% of the 5000 default), and the loop re-checks `reads >= remReads` at the top so it never starts another batch (non-runaway). `read_limit` is an internal cost guard, not a client contract. Budgeting ~2 reads/snapshot doesn't improve the worst case: under heavy `remote_endpoint` post-filtering the state docs are read regardless of whether they pass the filter, and a smaller batch just adds round-trips. This was analyzed in the spec's Round-2 performance review (resolved: document, don't re-engineer) and the code carries a NOTE at the `HISTORY_BATCH` constant stating the effective ceiling is `read_limit + HISTORY_BATCH`.

functions/src/api/bulk-read.ts
```diff
@@ -0,0 +1,274 @@
...
+      items.push(...r.items);
+      reads += r.reads;
+      bytesUsed += r.bytes;
+      if (r.touched) { touched.push(r.touched); }
+      endpointsWalked++;
+      stopOffset = offset;
+      stopCursor = r.innerCursor;
+      stopExhausted = r.exhausted;
+
+      if (!r.exhausted) { break; }                     // hit a cap mid-endpoint -> resume same endpoint
```
emcelroy, line 257: Claude suggestion:

**Stop walking endpoints once the response-byte budget is reached.**

Each endpoint admits its first item regardless of `remBytes`. Because Line 256-259 never checks `bytesUsed`, many endpoints can cumulatively exceed the 8 MB budget and cause response delivery to fail.

Add a regression covering multiple endpoints whose aggregate payload crosses the budget.

```suggestion
      if (bytesUsed >= RESPONSE_BYTE_BUDGET) { break; }
      if (!r.exhausted) { break; }                     // hit a cap mid-endpoint -> resume same endpoint
```

server/lib/report_server_web/api/v1/bulk_export_controller.ex
```diff
@@ -0,0 +1,181 @@
...
+  end
+
+  # ---- page 1: derive-once, mint scratch + intent row atomically, serve from index 0 / null cursor ----
+  defp first_page(conn, user, report_run, data_type, collection, limit) do
+    case EndpointSet.derive_endpoint_set(user, report_run) do
+      {:ok, []} ->
+        # empty export: terminal empty page, no scratch to resume — but still record the intent row for audit
+        # completeness, fail-closed like every other audit write. export_id is minted so the row is correlatable.
+        export_id = Exports.mint_scratch_id()
+        intent_attrs = audit_attrs(user, report_run, "export_scoped", "export_scoped", export_id, nil, [])
```
emcelroy: Claude suggestion:

**Record the actual export type on intent rows.**

Both calls store "export_scoped" as `data_type`, making answers and history intents indistinguishable and breaking data-type audit filtering. Keep that value as the event, but pass `data_type` to the fourth argument.

Add a test asserting the intent row's `data_type` matches the route (the current suite would happily pass with the value wrong, which is probably how this slipped through — visible in the empty-export test, which checks `event` and `endpoint_set` but not `data_type`).

```suggestion
        intent_attrs = audit_attrs(user, report_run, "export_scoped", data_type, export_id, nil, [])
```

functions/src/index.ts
```diff
@@ -56,13 +61,15 @@ api.get("/resource", getResource)
 api.get("/answer", getAnswer)
 api.get("/plugin_states", getPluginStates)
 api.get("/student_feedback_metadata", getStudentFeedbackMetadata)
+api.post("/bulk_read", requireHeaderBearer, bulkRead)
+api.post("/fetch_attachment_meta", requireHeaderBearer, fetchAttachmentMeta)
 // TODO: comment out for final PR
 api.get("/fakeAnswer", fakeAnswer)
```
emcelroy: I'm thinking there are more PRs coming, but just want to flag this in case it was overlooked for this one?

emcelroy (commented):
Looks good 👍 I'm approving although you may want to address at least the F1 issue in the review below generated with Claude, and remove lines 66-67 of `functions/src/index.ts` (unless this isn't the final PR, of course).

## In-depth review notes

Overall this is high-quality, carefully-reasoned work. The authorization model is sound and matches the revised spec, the cursor/pagination logic is meticulous (the gap/dup reasoning at cap boundaries checks out), audit writes are genuinely fail-closed, and the defensive validation against direct-to-Node calls is thorough. One correctness bug worth landing before real large-run testing, plus some minor/hardening notes. None of the below is blocking except as noted on F1.

### 🔴 F1 — Byte-budget overflow can *hard-stall* an export 
Adding to the inline note about checking `bytesUsed` between endpoints: the failure mode is worse than a single bad page. The oversized response is deterministic for a given token (same slice, same data), so:
- A page that produces >10 MB → gen1 rejects the oversized response → Elixir sees a non-200/transport error → `server_error` (500), no cursor advance.
- The CLI retries the same token → reproduces the same >10 MB response → 500 again. Restarting from a null token re-derives the same endpoint set and stalls at the same offset.
- Net: for affected data shapes the export is **unrecoverable** until fixed — a permanent stall, not graceful degradation.

Reachability is real for this feature's target data: history full-state snapshots are large, so the 8 MB byte budget (not the 500-item cap) is the binding constraint, and a chain of single-item/exhausted endpoints past the 8 MB point accumulates unbounded. Recommend landing the `if (bytesUsed >= RESPONSE_BYTE_BUDGET) break;` guard (plus a regression seeding multiple exhausted single-item endpoints whose aggregate payload crosses the budget) before exercising `/answers` + `/history` on real large runs.

### 🟡 F2 — Authorization is snapshot-at-start; mid-export project-permission revocation is not live (by design — confirming intent)
The Jira story says "re-check allowed_project_ids EVERY page," but the closed spec deliberately reversed this (spec §"Snapshot-at-start authorization" and the mid-export-change list): the endpoint set is frozen at export start, and per-page liveness is limited to (a) API-token revocation and (b) local `portal_is_*` role flags, both enforced live by `AuthPlug` on every request. A pure project-permission revocation only takes effect on the next export (or after the ≤1h TTL). The code correctly implements the spec — flagging only to consciously sign off on that tradeoff.

### 🟢 F3 — `--inline` disposition passes an unvalidated `contentType` to the browser (low)
The inline branch in `server/lib/report_server_web/aws.ex` sets `response-content-type` to the doc's `contentType` with no validation. A student-supplied `contentType: "text/html"` would make S3 serve the attachment as HTML, so a researcher opening it inline renders arbitrary markup. Low severity — opt-in (`--inline`), executes in the S3 presigned-URL origin (not the app origin), and concerns the researcher's own authorized students' content. Worth a small hardening: whitelist a safe content-type set for inline, or fall back to `application/octet-stream` otherwise.

### 🟢 F4 — `fakeAnswer` dev endpoint writes arbitrary Firestore docs (pre-existing, not this PR)
`functions/src/api/fake-answer.ts` does `getDoc(path).set(answer)` with fields spread from query params, wired into the live API with a `// TODO: comment out for final PR` above it. It predates this branch (last modified in a non-REPORT-76 commit; the route shows as unchanged context in the diff), so it's not a regression here and shouldn't block. Flagging since the TODO raises the question of whether this is the "final PR" — probably best cleaned up separately.

### ⚪ F5 — Minor notes (very low)
- Per-page `bulk_read` access rows log only endpoints that returned items; an endpoint that's read but yields zero items appears only in the intent row. Consistent with "data that actually left" semantics — noting it's a deliberate granularity choice.
- The admin `remote_endpoint` audit filter is a pathless `JSON_CONTAINS` full scan — fine for admin-only use now, could get slow on a large audit table later.
- **Response-semantics validation** in the bulk-read success clause: offset-range check + `is_boolean(exhausted)` (guards silent truncation / silent mis-branch). Skip the `next_inner` part — it's redundant with the existing per-page cursor validation.
- **Atomic TTL renew** in `bump_ttl` (renew with an `expires_at > now` predicate, treat a zero-row update as `:expired`): closes a narrow TOCTOU and removes a latent `StaleEntryError` 500 on the history path.

functions/src/index.ts
dougmartin: That has been in there for a few years - I'll remove it :)

functions/src/api/bulk-read.ts
dougmartin: Fixed in e0655d5 — landed the suggested `if (bytesUsed >= RESPONSE_BYTE_BUDGET) { break; }` guard between endpoints, plus the regression test: 10 exhausted single-item endpoints of ~1 MB each, where only the walk-loop check (not the per-endpoint trim) can stop the chain. The walk now stops once the budget is spent and the next page resumes from the following endpoint with no gap/dup. Verified the test fails with the guard removed.

server/lib/report_server_web/api/v1/bulk_export_controller.ex
dougmartin: Fixed in e0655d5 — both intent rows now pass the route's `data_type` as the fourth argument, so answers and history intents are distinguishable in audit filtering. Also added the missing assertions: the empty-export and page-1 audit tests now check the intent row's `data_type`, and a new test covers the history route.

functions/src/index.ts
dougmartin: Removed in e0655d5 — the route, its import, and `fake-answer.ts` itself are gone.

dougmartin:
Thanks for the thorough review — everything actionable landed in e0655d5:

**F1** — fixed; see the inline thread (byte-budget break between endpoints + regression test).

**F2** — confirming the intent: snapshot-at-start is deliberate per the closed spec. Per-page liveness stays limited to token revocation and the `portal_is_*` role flags (both enforced by `AuthPlug` every request); a pure project-permission revocation takes effect on the next export or after the ≤1h TTL.

**F3** — hardened: inline presigns now whitelist the response content type (`image/`, `audio/`, `video/` prefixes with `image/svg+xml` excluded, plus PDF/JSON/plain/CSV); anything script-capable like `text/html` degrades to `application/octet-stream`. Covered by a new test.

**F4** — the `fakeAnswer` endpoint is removed (route, import, and handler); see the inline thread.

**F5** — both actionable notes landed: the bulk-read success clause now validates response semantics (offset in range, boolean `exhausted`, list `items`) and fails closed on violation, skipping the redundant `next_inner` check as suggested; and `bump_ttl` renews atomically with an `expires_at > now` predicate, treating a zero-row update as `:expired`. The two observational notes (zero-item endpoints appear only in the intent row; pathless `JSON_CONTAINS` scan on the admin filter) are acknowledged as deliberate for now.

PR #397, opened 2026-07-17
cc-data CLI server support: token self-management, exchange label, export coverage [REPORT-77]

dougmartin:
**Jira**: https://concord-consortium.atlassian.net/browse/REPORT-77
**Spec**: [specs/REPORT-77-cli-server-support.md](https://github.com/concord-consortium/report-service/blob/REPORT-77-cli-server-support/specs/REPORT-77-cli-server-support.md)

Server-side pieces owed by the cc-data CLI story. All three are small additive changes to API surface that shipped in REPORT-74/76; the CLI degrades gracefully when any piece is missing, so shipping order between the repos does not matter.

## Token self-management: `GET` and `DELETE /api/v1/tokens/current`

- `DELETE` revokes the calling bearer token (backs `cc-data logout`), returning 200 `{"revoked": true}`; a lost concurrent race is the same success, and any reuse of the revoked token gets the standard 401.
- `GET` returns the calling token's own metadata — `label`, `created_at`, `last_used_at`, `report_access` (backs `cc-data auth status --check`).
- Both routes run in a new token-only `AuthPlug` mode (`token_only: true`) in a dedicated pipeline: the bearer is verified but the `can_access_reports?` role gate is skipped, so a de-provisioned user can still inspect and revoke their own credential, and `touch_api_token` is skipped, so a status check never overwrites the `last_used_at` signal it reports. `report_access` carries the role-gate result as data instead.
- The new scope sits above the `/api/v1` catch-all fallback scope, which would otherwise shadow it (Phoenix matches in definition order and the fallback matches every verb).
- `AuthPlug` now assigns `:api_token` on every authenticated route; existing routes keep their exact prior behavior.

## Optional `label` on the CLI token exchange

`POST /auth/cli/token` accepts an optional `label` string that replaces the constant "CLI login" label (the CLI sends `CLI login (<hostname>)`), so CLI-minted tokens are distinguishable in the token UI. Sanitized rather than rejected — trim, truncate to 100 chars, trim again, empty/non-string falls back to the default — because the exchange burns the one-time code, so a 400 over a cosmetic field would cost the user a whole login round-trip.

## `total_endpoints` on the answers/history export envelope

Both bulk-export envelope emission points gain `total_endpoints`: the size of the endpoint set snapshotted at export start (`0` on the empty-export page), constant across all pages of one export, so the CLI can record honest "queried vs. with data" coverage counts. Additive; `GET /api/v1/reports` keeps its two-key envelope.

## `report_type` on run metadata (added post-open)

A fourth owed dependency from the cc-data-cli spec review, recorded as a dated amendment in the closed spec: the run JSON (both list items and show) gains an additive string field `report_type` with a server-owned vocabulary of `answers` | `usage` | `log` (`student-answers` → `answers`; `student-assignment-usage` → `usage`; the three log-based reports → `log`). The CLI keys its pseudo-header row filter and report-shape allowlist on it and quarantines unrecognized values.

The value is declared on the Athena report modules (`api_report_type` in the `use ReportServer.Reports.Report` opts) and read from the report tree by the JSON view — derived at render time from `report_slug`, so all existing runs carry it with no migration. A slug-driven controller test asserts the exact value per report slug in both payloads and fails if a future Athena report ships without a vocabulary value (the vocabulary is API contract; additions are contract changes).

## Testing

- 405 tests, 0 failures (28 new tests across the branch, plus new `TokenControllerTest` and Athena `UploadTest` files).
- The unreachable-over-HTTP revoke race branch (`{:error, :already_revoked}` → 200) is covered by invoking the controller action directly with a pre-revoked token struct in assigns.

## Additional fixes from staging testing

Found while exercising the branch against staging data:

- **Fail closed on Athena staging uploads**: `LearnerData.upload/1` and `ResourceData.upload/1` discarded S3 put results, so a lost learner/resource staging upload produced a plausible-looking *empty* report instead of a failed run. Any put failure now fails the run; `AthenaDB` is env-injected for testability.
- **Nil `report_filter` crash**: runs created programmatically (API/console/fixtures) store a NULL `report_filter`, which crashed the my-runs/all-runs tables and the run show page. The shared component now normalizes nil to the empty filter, mirroring `ReportJSON`.
- **Functions lockfile** (REPORT-76 rider): `firebase-tools` was added to `functions/package.json` without regenerating `package-lock.json`, breaking `npm ci` in every cloud build — no function could deploy. Lock resynced.
- Minor: dev default Athena log DB corrected to `log_ingester_qa`; reports-landing token/audit links regrouped and renamed (Your/All API Tokens); cli-token intro paragraph spans full width.

### DIFF
--- a/functions/package-lock.json
+++ b/functions/package-lock.json
@@ -26,6 +26,7 @@
         "@types/jest": "^24.0.15",
         "@types/node": "^22.0.0",
         "firebase-functions-test": "^0.3.3",
+        "firebase-tools": "^15.0.0",
         "jest": "^24.8.0",
         "ts-jest": "^24.0.2",
         "tslint": "^5.12.0",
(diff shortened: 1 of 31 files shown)


emcelroy (approved):
Looks good 👍

PR #399, opened 2026-07-26
docs: call out the separate firestore rules deploy step [REPORT-84]

dougmartin:
Docs only. **Not required for the Functions 1.7.0 release** — this can merge whenever, independently of REPORT-84.

## Why

`npm run deploy` in `functions/` resolves to `firebase deploy --only functions`, so it does not carry the firestore rules. Rules deployment was already documented under `Deploying rules`, but neither deploy checklist pointed at it, which makes it easy to skip during a release.

The failure mode is unpleasant to diagnose: the functions deploy succeeds with no error, and the problem surfaces later as a `permission-denied` in the client. On the chat tutor work that reads as a generic "tutor unavailable", which looks like a functions bug rather than a missing deploy step.

## Changes

- Cross-link the rules deploy from both the dev and production checklists, conditional on `../firestore.rules` having changed.
- Add a callout under the production checklist explaining that `npm run deploy` is functions-only.
- Note on `Deploying rules` that it targets whatever `firebase use` currently selects, and that nothing in CI runs it.
- Add an `Indexes` section describing the tracked-index workflow: the file is the source of truth for a deploy (it creates what it lists and proposes deleting what it does not), a new query shape needs its index added to the file, and clicking the console link from a `failed-precondition` error only creates the index in the project you were running against, which is how the two projects drift apart.

No behavior changes, no code touched.

## Depends on #401

The `Indexes` section describes `firestore.indexes.json` as populated, which #401 makes true. Merge that first, or this ships documentation describing a file that is still empty.

### DIFF
--- a/functions/README.md
+++ b/functions/README.md
@@ -132,7 +132,7 @@ The deploy will fail with clear instructions if any required secrets are missing
 * set the current project to production: `firebase use report-service-pro`
 * deploy the functions: `npm run deploy` (this generates the build info)
 * if `../firestore.rules` changed, also deploy the rules (see [Deploying rules](#deploying-rules))
-* if the release adds a new query shape, exercise it in production and create any index it asks for (see [Indexes](#indexes))
+* if the release adds a new query shape, deploy its index too: `firebase deploy --only firestore:indexes` (see [Indexes](#indexes))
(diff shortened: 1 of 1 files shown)


emcelroy (approved):
Looks good 👍

PR #400, opened 2026-07-26
Functions 1.7.0: log-first authoring host, object_store rules, version bump

dougmartin:
Three related commits that together make up the Functions 1.7.0 release.

## Commits

**`67bc199` restore object_store rules that existed only in the deployed rulesets**

`object_store_data` and `object_store_metadata` have never been in `firestore.rules` — `git log -S object_store_data --all` finds nothing. They were added through the Firebase console: dev on 2025-12-02 (ruleset `3c995966`), pro on 2026-01-21 (ruleset `2671e419`).

Because the repo never carried them, the rules deploy on 2026-07-06 (ruleset `f3c6420e`, the chat tutor work) silently dropped both blocks from staging. Production kept them only because it had no rules deploy since 2026-04-03. So a rules deploy from this repo was a **regression** for production, not an update: it would have revoked create/read/update on two collections holding live data (`object_store_metadata` under `sources/activity-player.concord.org`, most recent doc 2026-03-30, two READY composite indexes).

Restoring them makes the file a superset of both deployed rulesets. Verified with a whitespace-insensitive diff against the live production ruleset: zero removals.

**`cc34091` derive the log-first authoring host from per-project config**

A log-first turn (forwarded telemetry arriving before any typed message) has no client `activityUrl`, so `composePageSystemPrompt` fell back to `defaultActivityUrl`, which hardcoded `AUTHORING_HOSTS[0]` (`authoring.concord.org`) regardless of deployment project.

On `report-service-dev` that fetched **production** authoring for every log-first turn, where the same activity id is an unrelated activity — so `findPage` threw `no visible page matches pageId N`, the drain died, and the parent doc went to `status:"error"`, surfacing in AP as "The tutor is currently unavailable."

`defaultActivityUrl` now takes an optional host from a new `AUTHORING_HOST` param (same pattern as `OPENAI_MODEL`), set per-project in the `.env` files. The host is still checked against `AUTHORING_HOSTS`, so a mistyped `.env` falls back to the default rather than widening the fetch target — SSRF posture unchanged, still no client-controlled input on this path.

Production was unaffected by the bug: its correct host happened to be the hardcoded one.

**`6a8a2bd` functions 1.7.0 and OPENAI_MODEL for production**

Bumps 1.5.0 → 1.7.0 (skipping 1.6.0, which tagged without ever bumping this file). Also sets `OPENAI_MODEL` for `report-service-pro`, which was only ever set for dev; as a `defineString` with no default it would have returned `""` in production and failed every chat turn silently.

`OPENAI_API_KEY` was likewise absent from `report-service-pro` and was created there separately, outside the repo, along with its `secretAccessor` binding for the App Engine service account.

## Verification

Staging: fresh authenticated conversation drains 19 forwarded log docs with no function errors, replies grounded in the correct activity. `object_store` rules restored (they had been missing for three weeks).

Production: `chatTutorOnWrite` created, rules deployed (additive — `object_store` preserved), both `messages` composite indexes created, authenticated learner on a real production activity got a correctly grounded reply. cc-data-cli verified separately against the `api` function with no 4xx/5xx.

Tests: 6 chat suites, 60 passed / 4 skipped. Lint and `tsc` clean.

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -135,6 +135,18 @@ service cloud.firestore {
       allow update: if studentWorkUpdate();
     }
 
+    match /sources/{source}/object_store_data/{document=**} {
+      allow create: if studentWorkCreate();
+      allow read: if studentWorkRead();
+      allow update: if studentWorkUpdate();
+    }
+
+    match /sources/{source}/object_store_metadata/{document=**} {
+      allow create: if studentWorkCreate();
+      allow read: if studentWorkRead();
+      allow update: if studentWorkUpdate();
+    }
+
     match /sources/{source}/plugin_states/{document=**} {
       allow create: if studentWorkCreate();
       allow read: if studentWorkRead();
--- a/functions/.env.report-service-dev
+++ b/functions/.env.report-service-dev
@@ -1,2 +1,3 @@
 AWS_S3_BUCKET=concord-staging-report-data
 OPENAI_MODEL=gpt-5.5
+AUTHORING_HOST=authoring.lara.staging.concord.org
--- a/functions/.env.report-service-pro
+++ b/functions/.env.report-service-pro
@@ -1,1 +1,2 @@
 AWS_S3_BUCKET=concord-report-data
+AUTHORING_HOST=authoring.concord.org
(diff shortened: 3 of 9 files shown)


dougmartin:
Merged without review as to not block the weekend release.

PR #405, opened 2026-07-29
feat: class lookup by word and the enroll-into-class step [REPORT-79]

dougmartin:
Implements [REPORT-79](https://concord-consortium.atlassian.net/browse/REPORT-79), the first of the fall-2026 "I'm Done" building blocks. Spec: `specs/REPORT-79-class-lookup-and-enroll.md` (the spec was closed on this branch: the per-spec folder was replaced by that single summary, which carries the finalized requirements, what was deliberately left unimplemented, and every decision with its options and rationale; the pass-by-pass `requirements.md` / `implementation.md` remain in git history).

Portal class ids differ between staging and production, so today's spring button hardcodes them and does not travel between environments. This branch adds the shared class-lookup-by-word read and a pipeline step that enrolls a student into a class named by its environment-stable class word, resolved to a `clazz_id` at run time. There is no user-facing behavior change yet: the new step is deliberately **not** wired into any `PIPELINES` entry (REPORT-82 owns that), so nothing in the shipped spring pipeline reaches it.

## What's here

**A typed step handoff.** `StepResult` gains `output?: StepOutput` (`{ destinationClassWord?: string }`). Unlike `summary`/`message`, which `send-email` renders into the teacher-notification email body, `output` is never rendered into a human-facing sink, so it is the only safe carrier for a value a later step consumes. REPORT-81's randomization step populates it; a regression test in `send-email.test.ts` asserts a step's `output` never reaches the email.

**`functions/src/tasks/portal-reads.ts`** (new), the layer above `portal-api`'s transport: `lookupClassByWord` (`classes#info?class_word=`, returning the class, its teachers, and its offerings) and `resolveOriginOffering` (`offerings#show`, returning `clazz_id` and the teacher-gated `class_word` from one response). Both are pure reads that take an already-minted token and return `{ status }` on failure, leaving minting and classification to the calling step. Two details worth a reviewer's eye: the offering mapper keeps `url` (the offering's own API url) and `external_url` (the activity url) distinct, since REPORT-80 matches on the latter, and `PortalTeacher` carries a PII guard because `get_info` anonymizes students but returns teachers' real names.

**`send-email` now reads its origin offering through that helper**, as its own behavior-preserving commit. The only difference on the wire is dropping the `reason: data?.details?.reason` argument to `classifyPortalFailure`, which `offerings#show` never returns, so every failure keeps its bucket and the existing tests stay green.

**The `enroll-specified-class` step.** It resolves its destination word from either the authored `target_class_word` param or a prior step's `output.destinationClassWord`, looks that word up with an origin (unscoped) teacher token, mints a destination-scoped cross-class token, and posts `add_to_class`. Idempotency rests on the portal's server-side no-op, so a re-run for an already-enrolled student succeeds. One decision worth calling out: an authored word that **differs** from a present handoff is a hard configuration error (tell-your-teacher, logged at error, no portal calls) rather than a silent preference, because `target_class_word` is a per-launch param, so silently preferring it would route every student to one class and defeat randomization study-wide. An authored word equal to the handoff is not a conflict.

**Harness coverage.** The stub's `classes/info` is now keyed on the requested `class_word` and serves an origin class and a destination class with distinct ids and names, 400s an unknown word the way rigse's `classes#info` does, and its offering fixtures gain the `url` / `external_url` fields real `get_info` always returns. A new `run-step.js` drives one compiled step directly against the stub (no emulator, no seed, but a prior `npm run build`), twice against a single `StepContext`. The enroll assertion is on the summary's class name, which flows from resolving the destination word, so a wrong-class resolution fails the run instead of riding the stub's unconditional `{ success: true }`.

## Security notes

Both new portal call sites use `StepContext.portalOrigin` (the normalized origin `validatePortalHost` produced) and never the raw `platform_id`; the step's header comment records that it makes no host check of its own and that the consuming pipeline must keep the setup gate. Unit tests assert that across success and every failure path, neither a minted token value nor a real teacher name from the looked-up class appears in any log call or on the returned `StepResult`.

## Verification

`tsc`, `tslint`, and `jest src` (18 suites, 301 tests) are green. The full local harness passes 24/24 scenarios, which includes the spring pipeline running end to end against the refactored `send-email`. The stub log confirms the enroll posts the destination `clazz_id` (30001, not the origin's 90210) and that the second run of the step issues zero mints, both scoped tokens coming from the run's cache.

The closed spec's final decision entry records every place the shipped code diverges from the plan: four local renames tslint's `no-shadowed-variable` forced, the whole-module `portal-api` mock in the `portal-reads` tests, the scenario `driver` field and `run-all.js` dispatch the second harness driver needed, the class fixtures moving into `config.js` with a `classes` behavior key, and the failure-side assertion in `run-step.js`.

[REPORT-79]: https://concord-consortium.atlassian.net/browse/REPORT-79?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emcelroy (approved):
Looks good to me 👍 The review below was generated with Claude code. Two issues are raised but neither is blocking.

## Overall

The portal-read layer is a sensible boundary, the step uses the normalized `portalOrigin` consistently, the origin and cross-class token scopes are correct, and the new structured output stays out of the teacher email. The lookup and enroll failure buckets match REPORT-83, the destination class word is URL-encoded, and the step does not put the lookup's teacher data into logs or its `StepResult`.

Below are one small hardening fix worth taking now and one design question worth settling before REPORT-82 wires the step into a live pipeline.

## 1. Suggested fix — do not accept a malformed 2xx class response as a resolved class

**File**: `functions/src/tasks/portal-reads.ts:63`

The lookup considers any 2xx body with `id !== undefined` successful, which admits `id: null`:

```ts
const ok = response.status >= 200 && response.status < 300 && response.data?.id !== undefined;
```

A 200 carrying `{ id: null }` reaches the enroll step, which does `String(lookup.class.id)` and so mints a destination token for class id `"null"` and posts `clazz_id: "null"`. If `name` is also absent, the success summary becomes `Enrolled in undefined`. Missing or wrong-typed `name` and `class_word` escape the helper's type boundary the same way.

What makes this clearly worth fixing rather than a matter of taste: the sibling helper in the same file already guards the null case. `resolveOriginOffering` at `portal-reads.ts:117-118` does

```ts
clazzId !== undefined && clazzId !== null
```

so the two functions written together disagree about what a valid 2xx body is. Aligning `lookupClassByWord` with its neighbor is the smallest change that removes the inconsistency.

Worth validating, in rough order of payoff:

- `id` is a non-null number or nonblank string (this is the one that can misroute an enrollment);
- `name` is a nonblank string (this one reaches the teacher email);
- `class_word` is a nonblank string.

`teachers` and `offerings` need no new guard — the mapper at `portal-reads.ts:73-85` already coerces a non-array to `[]` on both. Returning `{ status }` for a malformed success body sends the caller down the existing classified generic-failure path with no other change.

Severity is low in practice: rigse's `classes#info` returns the class or a 400, and the harness stub models that 400 faithfully, so a 200 with a null id is not a shape the portal produces. This is type-boundary hardening, not a live defect.

## 2. Design question — should conflicting handoffs be an error?

**File**: `functions/src/tasks/ai4vs-flvs/enroll-specified-class.ts:33-40`

`resolveDestinationWord` stops at the first nonblank `output.destinationClassWord` in `stepResults`. Because `stepResults` is keyed by step name at `index.ts:81` (non-integer keys, so `Object.values` yields insertion order), that is reliably pipeline order — if two preceding steps ever produced different words, the earlier step would win silently.

To be clear, this is a deliberate, documented choice, not an oversight: the header comment states the invariant, and `enroll-specified-class.test.ts:149` ("takes the first prior step that set a destination word") pins the behavior with two *differing* words. So the question here is whether to revisit the decision, not whether the code matches its intent.

The argument for revisiting is consistency with the rule the step already commits to one branch earlier. An authored word that differs from a handoff is a hard `TELL_TEACHER_MESSAGE` error precisely because a config mistake that lands a student in a valid-but-unintended class is invisible — the lookup, mint, and enroll all succeed. Two handoffs that disagree have the same consequence and the same invisibility, but are currently resolved by position. Collecting the distinct nonblank handoff values and failing when there is more than one (equal duplicates staying valid) would make the two cases behave alike. Alternatively, name the producer step in the contract and read that result explicitly.

The argument against is that the exposure is narrow. `StepOutput` has a single field, only REPORT-81's randomization step is intended to populate it, and reaching this state requires someone to deliberately wire a second producer into a pipeline. It is a guard against a future authoring error rather than a latent bug in what ships here.

If you do take it, note the cost is not purely additive: `enroll-specified-class.test.ts:149` asserts the current behavior and would need to be rewritten alongside the header comment, not just supplemented with new cases.

dougmartin:
Thanks Ethan, both addressed in aa15607. Took the first, declined the second but wrote the reasoning into the code so it does not get re-raised as an oversight.

## 1. Malformed 2xx bodies: fixed

`lookupClassByWord` now validates the fields the callers actually consume before returning a class, and reports a malformed success body as `{ status }` so the caller takes the same classified-failure path as any other unusable response:

- `id`: finite number or nonblank string
- `name`: nonblank string
- `class_word`: nonblank string

`teachers` and `offerings` are left unguarded, as you noted the mapper already coerces a non-array to `[]` for both.

One thing I did beyond the suggestion: `PortalClass.id` was typed `number` while `OriginOffering.clazzId` is `number | string`, so I widened it to match. The substance of your point was that two functions written together disagreed about what a valid identifier is, and the type declarations disagreed too, so aligning only the runtime check would have left half the inconsistency in place.

Nine tests added: eight rejected shapes (null id, non-finite id, blank string id, and missing / blank / non-string `name` and `class_word`), plus one confirming a numeric-string id still resolves.

Agreed on your severity read. rigse's `classes#info` answers an unknown word with a 400 and the stub models that faithfully, so this is type-boundary hardening rather than a live defect. The comment says so, so nobody later reads the guard as evidence the portal returns null ids.

## 2. Conflicting handoffs: keeping the current behaviour, with the rationale recorded

Behaviour is unchanged, so `enroll-specified-class.test.ts:149` stands as written.

You framed this exactly right as "whether to revisit the decision", and the answer I landed on is that the asymmetry is justified rather than accidental, which is what the header comment now says. The two cases look symmetric but differ in who can cause them:

- An authored `target_class_word` is typed by a human into a `taskParams` blob. That is precisely the surface where a typo produces a valid-but-unintended class with the lookup, mint and enroll all succeeding, which is the invisibility you described. Worth hard-failing on.
- A second step publishing a different `destinationClassWord` requires a developer to wire a second producer into a pipeline. That is our own code, and it fails on the first harness run of that stage.

The standing principle we have been applying is that our own pipeline wiring is assumed correct rather than checked with defensive run-time code. It is the same reason the REPORT-81 spec dropped a guarded accessor that would have verified step ordering before reading `originClassWord` from `stepResults`. Adding a multi-handoff check here would be inconsistent with that in the other direction.

The comment records the decision, the reasoning, and an explicit "revisit if a second producer of this field is ever intended", so the next reader finds the argument rather than re-deriving it.

## One correction to a premise, which does not change the conclusion

Your "exposure is narrow" argument partly rests on `StepOutput` having a single field. REPORT-81 adds `originClassWord`, published by the new `resolve-origin-class` step, so it will not. Only one step still produces `destinationClassWord`, so the conclusion holds, but I restated the invariant in the comment as **one producer per field** rather than one field on `StepOutput`, since the old wording would have read as stale the moment REPORT-81 lands.

Checks: `tsc` clean, `npm run lint` clean, 224 tests passing across all eight suites in `src/tasks`.

PR #406, opened 2026-07-30
Pilot-configurable randomization for the fall study [REPORT-81]

dougmartin:
Implements [REPORT-81](https://concord-consortium.atlassian.net/browse/REPORT-81). Spec: [specs/REPORT-81-pilot-configurable-randomization.md](specs/REPORT-81-pilot-configurable-randomization.md).

Makes the "I'm Done" randomization step select its strata table, its demographic inputs and its balancing pool from the student's **program**, resolved at run time from their class word, so one shared fall pre-test button randomizes both cohorts correctly. Also delivers the step that resolves the class word in the first place, which REPORT-80 and REPORT-82 both need.

Neither new step is wired into `PIPELINES` and `index.ts` is untouched: REPORT-82 owns the fall stages, their pilot keys and the step order. This PR adds steps; it does not wire them into a pilot.

## What changed

Eleven commits, each reviewable on its own. The first three work on the assignment document (a pure move, then a signature change, then the one behaviour change), so the reviewer of the behaviour change sees only the walk and its tests. The next two extract the demographic reading and then retarget its tests, split so the extraction lands with the existing suite untouched and green as its evidence.

- **`assignment-doc.ts`** — the document identity and the alternating transaction, moved verbatim. Its entire import list is `crypto` + `firebase-admin`, which is what lets the emulator test import real production code with zero mocking. `firebase-functions` pulls in `firebase-admin/auth` (jest 24 cannot resolve its subpath exports) and `firebase-client` pulls in `firebase/auth` (`fetch is not defined`), so adding a logger import here for one log line would silently cost the story its one unmockable proof.
- **`AssignmentScope`** — a `{ docId, fields }` pair the step computes and hands to the core, with `perClassScope` (byte-identical to the shipped key) and `pooledProgramScope`. The core gains no conditional and decides nothing.
- **Per-student de-duplication** — the one behaviour change, and the one deliberate change to spring.
- **`demographics.ts` / `pre-tests.ts` / `strata-tables.ts`** — extracted, with `random-assignment.ts` reduced from 450 lines to 146.
- **The 20-row full-time table**, transcribed in the source document's row order.
- **`resolve-origin-class`** and **`fall-random-assignment`**.
- **A `test_functions` CI job**, because nothing in CI has ever run this package.

## Three things worth a reviewer's attention

**The de-duplication is per student, not per (stratum, student), and spring inherits it.** The shipped lookup indexed straight into the current stratum, so a student whose stratum changed between two clicks was not found and was assigned again. That window is real: the pre-test is locked only *after* a successful run, so any failure between assignment and lock returns the student to an unlocked pre-test where they can edit an answer and click again. Because `add_to_class` only adds and this pipeline can remove nobody, the second assignment does not re-bucket the student, it enrols them into a second class alongside the first and contaminates **both** arms. That argument is about the enrolment primitive rather than about which cohort is running, so it applies to spring identically. `spring-2026` is dormant, so the blast radius is nil, and the changed case is pinned by a test rather than left as an incidental consequence of the extraction.

**The pooled flex key contains the program, and must.** `resource_link_id` is the offering id and an offering is per class per activity, so excluding only `context_id` would still yield three flex documents. Excluding `resource_link_id` as well leaves `interactiveId | platform_id`, and one shared Green button gives both programs the same `interactiveId`, so all eight classes would collapse into one document and full-time's within-teacher balancing would silently stop meaning anything, with every test still passing. The program id is year-qualified (`fall-2026-flex`) because, unlike the per-class key, nothing in the pooled key re-keys itself for a new cohort.

**Class words are lowercase, and the case-sensitive version was broken against real data.** `Portal::Clazz` runs `before_validation :class_word_lowercase` and `:class_word_strip` on every save, and `offerings#show` serves the stored value, so `FT-2026-Bingler` reaches this pipeline as `ft-2026-bingler`. Driven against the eight real class words as the portal stores them, the drafted case-sensitive classifier returned `undefined` for 8 of 8, which would have failed every fall student in both programs on launch day. `resolve-origin-class` now normalizes at the single producer, so "in the portal's stored form" is an invariant this story, REPORT-80 and REPORT-82 all inherit from one place. The harness stub was echoing mixed-case fixtures verbatim, which is the fidelity gap that would have let this pass locally; it now models the portal's lowercasing.

## Verification

- `npm run lint` and `tsc --noEmit` clean; **25 suites, 454 tests** (4 skipped, pre-existing).
- **The emulator suite is the load-bearing one.** The unit suite mocks `runTransaction` wholesale, so a mocked test can assert only that the de-duplication walk was *written* as intended, never that it behaves correctly inside a real transaction. `assignment-doc.emulator.test.ts` drives real production code against the Firestore emulator with no mocks at all: 7 cases covering the stratum change, the unadvanced counter, the read-only commit, cross-section pooling, per-class separation and a full 24-stratum walk.
- The three `run-step` harness scenarios pass before and after the fixture change.
- The transcription is pinned by all twenty `n1` values individually, the five block starts separately (which catches a whole teacher's block transcribed with inverted polarity), all four block boundaries, the 10/10 split and a surname-to-teacher bijection of exactly five.

## Still open, deliberately

Four pre-launch checks on `FALL_PRE_TEST`, listed in the closed spec. It ships as a copy of spring's wording on the PI's 2026-07-29 confirmation, pinned by a deep-equality test; if any check comes back different, the fix is one change that edits the object, updates its attribution comment and deletes the pin.

[REPORT-81]: https://concord-consortium.atlassian.net/browse/REPORT-81?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emcelroy (approved):
Looks good to me 👍 The report below was generated with Claude Code. I don't think any of the issues are blockers. I almost didn't even include #3 but left it in case you want to be thorough. I think #4 is probably a non-issue, too, but I've included it for consideration.

# REPORT-81 review — pilot-configurable randomization

Branch `REPORT-81-pilot-configurable-randomization`, reviewed against `master`, the
[committed spec](specs/REPORT-81-pilot-configurable-randomization.md) and the Jira story.

**Approve with changes.** Nothing here blocks the merge. Findings 1 and 2 should be settled before REPORT-82 wires these steps into a pilot; 3 and 4 are launch-readiness items rather than code changes.

---

## 1. Full-time double assignment can be cross-arm, contradicting the spec

[spec:135](specs/REPORT-81-pilot-configurable-randomization.md#L135) and the *Should the destination class word be persisted* decision both state that the one open double-enrolment path is same-arm, so no arm can be contaminated. That holds for flex and for spring. It does not hold for full-time, which is the path this story adds.

Full-time keeps the per-class key, which includes `resource_link_id` and `context_id`. Both change when a student changes class, so the walk at [assignment-doc.ts:158](functions/src/tasks/ai4vs-flvs/assignment-doc.ts#L158) is reading a different document and cannot find the student's existing arm. They are assigned again from the new teacher's counters, and the new arm can be the opposite one. Because `add_to_class` only adds, they finish enrolled in one teacher's `-gator` class and another teacher's `-shark` class. That is cross-arm contamination — what the walk exists to prevent.

Reachability is low but real. It needs a roster move between two full-time classes plus a re-completion of the Green pre-test. The re-completion is plausible: the answers query filters on `resource_link_id` and `context_id`, so the student has no answers in the new offering, and `lock-activity` locks per offering, so the new offering is unlocked.

**Do:** correct the spec paragraph and the decision entry to name the exception — same-arm for flex and spring, cross-arm for full-time, because that program's pool is the class the student left. Add one line to the walk's comment block, which argues the enrolment-primitive case as if it were scope-independent. Put a mid-study move between full-time classes on the study's operational list as something that needs the prior assignment inspected, not just a roster edit.

**Do not:** build a fix. Closing this properly means a program-wide student→arm index, which the story deliberately declined. Worth noting the flip side the PR could claim: pooling makes flex immune, because a section move lands in the same document.

## 2. Flex accepts any `fl-` class word; full-time rejects an unknown teacher

[fall-programs.ts:40](functions/src/tasks/ai4vs-flvs/fall-programs.ts#L40) classifies on `ft-` / `fl-` alone. Full-time then gets a second, strict gate: an unknown surname is a classified failure before anything is written. Flex has no equivalent. A mistyped or unexpected flex word — `fl-2026-sction1`, or the spring-era `fl-spring-2026-origin` the local harness still uses — classifies as flex, consumes an alternation slot in the pooled `fall-2026-flex` document, and derives a destination word that then fails at enrolment.

It self-heals once the class word is fixed, since the walk returns the original arm, so this is not a data-integrity finding. Two things still make it worth tightening. The story treats an unknown full-time teacher as a classified failure precisely so a misconfiguration cannot silently reach the assignment document, and flex gets the opposite treatment for the same class of fault. And the program ids are year-qualified on the explicit argument that a later cohort must not land in this study's document, while the classifier ignores the year — `fl-spring-2026-origin` and `fl-2026-section1` are indistinguishable to it.

**Do:** match `ft-2026-` / `fl-2026-` rather than `ft-` / `fl-`. That keeps no per-section list in code and makes the classifier as year-qualified as the ids it returns. Add `fl-spring-2026-origin` to the negative cases in `fall-programs.test.ts`; the existing negative case (`spring-2026-origin`) reads as coverage of this but is not, because the word the harness actually carries has the `fl-` prefix.

## 3. The pre-launch checklist is missing the destination class words

The four open `FALL_PRE_TEST` checks are the right ones. There is a fifth of the same kind that is not listed: nothing in this story or its tests establishes that the destination classes exist in the portal as `<origin>-gator` / `<origin>-shark`. The derivation at [fall-random-assignment.ts:175](functions/src/tasks/ai4vs-flvs/fall-random-assignment.ts#L175) is asserted against itself and against the spreadsheet convention, not against the portal.

If the spreadsheet's destination words differ in any way the derivation cannot reproduce — a plural, an abbreviation, a different separator — every fall student is randomized correctly, holds an arm, and fails at enrolment with tell-your-teacher. Same shape and blast radius as the gender-options check.

**Do:** add a fifth checklist row — resolve all sixteen derived words (eight origin classes × two arms) through the class lookup on the staging portal before launch.

## 4. The study's registration class words are in a public repository

The eight real origin words are new to the repo on this branch, in the spec and in two test files. `git grep` finds none of them on `master`. The repository is public and the branch is pushed.

A class word is the portal's student self-registration key — the PR's own stub comment says so. If self-registration is open on these classes, anyone with a learner account can join a study registration class, complete the Green pre-test and be randomized into an arm. The consequence is contaminated study data rather than exposed data.

The spec settled publicness for *teacher names* deliberately. It did not consider class words as a credential; they were classified as "authored, environment-stable, not PII", which is true and does not address self-registration.

**Do:** confirm self-registration is closed on the eight registration classes for the study's duration. That is the check that decides whether anything else is needed; rotating the words is the fallback if it cannot be closed. Note that whether an arbitrary learner can join cannot be established from this repository, and that the words are already published — using placeholder words in the spec and tests is worth doing for anything added later, but it retracts nothing from history.

---

## Smaller items

- **The program predicate is recomputed in four places.** `program === FULL_TIME_PROGRAM` appears at [fall-random-assignment.ts:88](functions/src/tasks/ai4vs-flvs/fall-random-assignment.ts#L88), [:118](functions/src/tasks/ai4vs-flvs/fall-random-assignment.ts#L118), [:129](functions/src/tasks/ai4vs-flvs/fall-random-assignment.ts#L129) and [:160](functions/src/tasks/ai4vs-flvs/fall-random-assignment.ts#L160), and all four must agree. The worst pairing is 88 and 118, where the dimension set requested and the dimension set checked are derived separately — drift between exactly those two is what the guard at 118 exists to catch. One `const isFullTime`, plus one `const dimensions` reused by the read and the guard, removes the coupling.
- **The flex table's type hides its own guard.** [strata-tables.ts:12](functions/src/tasks/ai4vs-flvs/strata-tables.ts#L12) declares `Record<string, Arm>`, so `TABLE[stratumKey]` is typed `Arm` for any string and the `if (!n1Assignment)` guards in both steps are dead code to the compiler. `Record<string, Arm | undefined>` makes them load-bearing at the type level and still narrows cleanly after the early return. The full-time lookup does not have this problem, since `Map.get` is already `| undefined`.
- **The 24-stratum emulator test has no explicit timeout.** [assignment-doc.emulator.test.ts:109](functions/src/tasks/ai4vs-flvs/assignment-doc.emulator.test.ts#L109) performs 25 sequential emulator transactions, with a `clearFirestore` `recursiveDelete` in `beforeEach`, against jest's 5s default (`jest ^24`, no `testTimeout` in the config). Pass `}, 30000)`. The emulator suite is this story's load-bearing proof, and a timeout flake there is what gets a suite marked unreliable and skipped.
- **The global `firebase-tools` install in the new CI job is redundant.** `functions/package.json` carries `firebase-tools` as a devDependency, the lockfile resolves it to 15.24.0, `npm ci` installs it, and `test:emulator` runs `npx firebase`, which prefers `node_modules/.bin/firebase`. The emulator version — and therefore the JDK 21 requirement the job's comment explains — comes from the lockfile, not from [line 39](.github/workflows/firestore-and-query-creator-tests.yml#L39). Drop that step; it buys the functions job nothing and implies it controls the version. The rules job still needs its global install, since it invokes `firebase` bare from `tests/`.
- **Enable the npm cache.** `actions/setup-node@v4` supports `cache: 'npm'` with `cache-dependency-path: functions/package-lock.json`, which cuts `npm ci` on every functions push.
- **The workflow's name no longer describes it.** `firestore-and-query-creator-tests.yml` / "Firestore and Query Creator Tests" now also runs the functions package. Rename it, or note the scope in the file.
- **A typo'd fixture override fails in the wrong place.** In [answer-doc-fixtures.ts:61](functions/src/tasks/ai4vs-flvs/answer-doc-fixtures.ts#L61), `genderIdMap[gender]`, `gradeIdMap[grade]` and `races.map(r => raceIdMap[r])` yield `undefined` for a misspelled override, surfacing as `Choice ID "undefined" not found` from inside `readDemographics` — production code blamed for a fixture typo. A small lookup helper that throws `unknown Gender choice "…"` points at the caller instead. Module's `?? "mO"` is the converse hazard: a misspelled module label silently becomes the `Other/not sure` choice, which is a valid answer, so the test passes while testing something else.

PR #408, opened 2026-07-30
feat: fall-2026 pipeline stages (pre-test, curriculum, post-test) [REPORT-82]

dougmartin:
Implements [REPORT-82](https://concord-consortium.atlassian.net/browse/REPORT-82): the fall-2026 study's three "I'm Done" pipeline stages, per [the closed spec](specs/REPORT-82-fall-2026-pipeline-stages.md).

## What this does

`PIPELINES` gains three entries, keyed by a pilot string that encodes the **stage**: `fall-2026-green` (pre-test), `fall-2026-blue` (curriculum), `fall-2026-orange` (post-test). The dispatcher is unchanged, no step is conditional, and each stage stays a flat fail-fast array.

The table is keyed by stage and never by program. Both cohorts run the same three stages, because the only program-dependent behaviour in the study lives inside `fallRandomAssignment`, which resolves the program from the origin class word itself. A program-shaped pilot such as `fall-2026-fulltime` is forbidden: one shared Green button serves both cohorts, so it would label every flex student as full-time.

Two ordering constraints are load-bearing rather than incidental. The pre-test locks **after** the enrol, so a failed enrolment leaves the student unlocked and able to re-click. The post-test locks **before** it opens, which is what makes `open-target-offering`'s "Your work has been saved" copy true by control flow rather than by hope; reordering those two requires that message to be revisited in the same change.

Two smaller pieces come along with it. `resolveOriginClass` now publishes the launch offering's `clazz_id` alongside its class word, from the same response, and `sendEmail` takes that handoff instead of issuing a second `GET /api/v1/offerings/:id`; it **keeps** its own read as the fallback, which is not dead code, since spring-2026 and the fall curriculum stage both reach it on every run. And the stale `fall-2026-fulltime` pilot is retired from all four places it appeared, three of which were shipped unit tests rather than the harness.

## Commits

Ten, each independently reviewable and each leaving both suites green.

| Commit | What |
|---|---|
| `324d56c` | Retire the stale `fall-2026-fulltime` pilot value |
| `1ccde0a` | Publish the origin clazz id and consume it in send-email |
| `5f7cbfa` | Wire the three fall-2026 pipeline stages |
| `4388ebf` | Add the fall class fixtures and launch contexts to the harness |
| `b985abd` | Serve a per-scenario class identity from the stub portal |
| `2ee2695` | Seed harness answers per scenario |
| `c52dfe8` | Make the harness driver's request, expectations and read-back per scenario |
| `ded654a` | Add the three end-to-end fall harness scenarios |
| `39a40af` | Document the harness's fall behaviour |
| `3c986b5` | Fail a harness scenario that declares no assignment expectation |
| `383d73e` | Close the spec |

The four harness plumbing commits add capability nothing uses until the scenarios land, which is deliberate: it keeps `run-all.js` green at every commit, and each still has a real acceptance check of its own rather than being reviewable only by reading.

## Verification

Everything below was run, not inferred.

- `npx jest`: **26 suites, 491 passed, 4 skipped (495 total)**, which is exactly the total the plan derived per step from a 484 baseline.
- `npm run lint` and `npm run build`: clean.
- The harness on the emulator against the stub portal: **31/31 scenarios**, confirmed over six consecutive runs (28 existing plus the three fall stages).
- The post-test run makes one mint, one `GET offerings/:id`, one `GET classes/info`, two `PUT`s and one `POST send_class_teachers`, with **no second offering read**. That absence is the clazz-id handoff working end to end, and the notification reaches the origin class the handoff carried.
- The two pre-test scenarios carry the **same** seeded answers and land in **opposite** arms (full-time → `ft-2026-bingler-gator`, flex → `fl-2026-section1-shark`). That is the point of the pair: identical demographics can only reach opposite arms if the resolved program actually selected a different strata table.

Three new assertions were checked to be falsifiable rather than vacuous, by breaking each and watching it fail: the ordered-handler test, the assignment document id formula, and the new declaration guard.

## Notes for the reviewer

**The one harness assertion that observes the pipeline.** The assignment read-back reads an *arm* out of the persisted document and recomposes the word from harness constants, so both sides of that comparison are harness values. Since every subclass word has a `classes/info` fixture on purpose, a pipeline that appended the **wrong suffix** would resolve a real class, enrol successfully, and pass. The stub therefore records each `add_to_class` body and the scenario asserts the `clazz_id` the pipeline actually enrolled into. The comparison is stringified on both sides because the two enrolling steps disagree on the type they post.

**`originClazzId` is a string on purpose.** `readStepOutputField` accepts nothing else, while `resolveOriginOffering` types `clazzId` as `number | string` and the portal serves a JSON number. Publishing the raw value would hand back `undefined` and `send-email` would fall back forever, silently, with nothing failing.

**The `index.test.ts` lock mock is keyed on the pilot, not the entry name.** One handler now runs under four different entry names, and the runner does not tell a handler which entry it is executing, so the pilot is the only thing in scope that distinguishes them.

**Five divergences from the plan as written** are recorded under "Implementation divergences" in the closed spec, so the spec and the code agree. The substantive one is the last commit: the driver's two read-back branches were opt-in, so a success scenario declaring neither `expect.assignedClassWord` nor `expect.noAssignment` reported PASS while verifying nothing beyond its completion text.

**Still outstanding, and not code.** R17's pre-launch checklist runs against the real fall classes once they exist. Every item is portal-side state the pipeline cannot establish for itself, and each failure is invisible until a student clicks and then affects a whole arm at once. The tickable copy is [a secret gist](https://gist.github.com/dougmartin/2b07c5b6794eebbcc0c743e94b06ec41), and it is also reproduced in full in the closed spec, which is where it survives the story.

---

The two spec files have been collapsed into a single closed summary at [`specs/REPORT-82-fall-2026-pipeline-stages.md`](specs/REPORT-82-fall-2026-pipeline-stages.md) and the source folder deleted, following the precedent of REPORT-81. The summary keeps the requirements as specified, what was deliberately left out and why, the implementation divergences, the R17 checklist, and the full decision record from four rounds of requirements self-review and one of implementation self-review.

[REPORT-82]: https://concord-consortium.atlassian.net/browse/REPORT-82?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emcelroy (approved):
LGTM 👍 The review below was generated with Claude and contains non-blocking issues to consider before merging.

## Overall

The wiring matches the spec and the two step changes are small and correct. Nothing below needs to hold the merge.

---

## 1. The post-test stage turns a portal misconfiguration into a "the code is mis-wired" message

**Files:** [open-target-offering.ts:130-137](functions/src/tasks/ai4vs-flvs/open-target-offering.ts#L130-L137), header at [:57-59](functions/src/tasks/ai4vs-flvs/open-target-offering.ts#L57-L59)

When the student's class word carries neither the `-gator` nor the `-shark` suffix, the open step fails with the shared message: *"Something went wrong setting up your class. Please tell your teacher."*

The header explains why that branch gets the shared message rather than the step's own: the fault "means the stage is mis-wired rather than the portal misconfigured," and "nothing about the student's work is knowable" from it. Wiring `fall-2026-orange` makes both of those false.

**The class word isn't ours.** `resolveOriginClass` publishes whatever `offerings#show` returns for the offering the student launched from. So an unclassifiable word means the Orange sequence is sitting in a class whose word has no arm suffix — a registration class, most likely. That's portal-side placement, the same category as "no offering matched the name" and "several offerings matched," both of which this same header deliberately routes to the step's *own* message. The test that pins this branch, [open-target-offering.test.ts:254-267](functions/src/tasks/ai4vs-flvs/open-target-offering.test.ts#L254-L267), even uses `"ft-2026-bingler"` as its example — which is an actual study registration class word — while calling it a mis-wired stage in a comment.

**And the student's work is now knowable.** In `fall-2026-orange` the lock runs before the open, so by the time this branch fires the post-test is locked and showing on the roster. The student is told something went wrong "setting up your class" — which describes neither what they were doing nor what failed — and is denied the "Your work has been saved" clause that is true for them. It fires for every control student in the affected class, at the last event in the study.

**Do:** route the unclassifiable-word branch to the step's own `STUDENT_FAILURE_MESSAGE`, update the two header paragraphs to say the class word is portal data now that the stage is wired, and flip the test's `not.toContain("Your work has been saved")` for that one case.

Leave the absent-handoff branch at [:123](functions/src/tasks/ai4vs-flvs/open-target-offering.ts#L123) alone. That one really is a wiring fault, and in the wired stage it's now unreachable anyway, since `resolveOriginClass` runs first and its failure stops the pipeline.

---

## 2. The pre-launch checklist doesn't check the assumption the post-test's arm read depends on

The spec states the assumption plainly at [spec:603](specs/REPORT-82-fall-2026-pipeline-stages.md#L603): the arm "is read from the class word the student launched from, which is only possible because Blue and Orange live exclusively in the subclasses."

R17's checklist covers Blue being *present* in every `-shark` class, the right initial locked/open state per arm, and all sixteen subclass words existing. It has no item for **exclusivity** — that Blue and Orange appear *only* in the subclasses, and Green *only* in the registration classes.

That's the assumption finding 1's branch fires on. It also matters in the other direction: a Green sequence added to a subclass would have `classifyFallProgram` see `ft-2026-bingler-shark`, [derive the teacher surname as `shark`](functions/src/tasks/ai4vs-flvs/fall-programs.ts#L83-L84), and fail the unknown-surname gate. For flex it's worse in a quieter way — the destination word becomes `fl-2026-section1-shark-shark`, an enrollment that can't succeed. Both fail loudly, but "loudly" here means a whole class of students blocked at their last click.

**Do:** add one item to R17, next to the class-state table since that table is already the PI's to own — confirm that Blue and Orange exist only in `-gator` / `-shark` classes, and that Green exists only in the registration classes. One query per sequence, and the cheapest item on the checklist to run.

---

## 3. The new failure log marks "the student hasn't finished yet" as an error

**File:** [index.ts:138-140](functions/src/tasks/ai4vs-flvs/index.ts#L138-L140)

The new line logs at `error` for every failed step. Most failures earn it. One doesn't.

[`evaluateCompletion`](functions/src/tasks/ai4vs-flvs/evaluate-completion.ts#L66-L77) returns failure when the student has answered fewer than `min_completed_questions`, with a "please answer more questions" message. It's not an error in any operational sense: nothing has been written, the student isn't locked, and clicking again after answering another question fixes it. Today that outcome logs only its counts at `info`. From this branch on, every early click on the Green button (and on spring's existing button) also produces an error-level line.

Worth being precise about scope, because it's narrower than it first looks. The other "student skipped something" outcome — a missing demographic answer, on both the fall and spring randomisation steps — **already** logs at error level, from [`demographics.ts:266-268`](functions/src/tasks/ai4vs-flvs/demographics.ts#L266-L268). This branch doesn't change that; it just adds a second line beside it. So the level question there is pre-existing, not new.

The line is worth keeping regardless. Its best case is one that logs nothing at all today: a Green button authored without `min_completed_questions` returns at [evaluate-completion.ts:28-32](functions/src/tasks/ai4vs-flvs/evaluate-completion.ts#L28-L32) with no log, so that misconfiguration was previously invisible outside the job document. The awkwardness is that it and an early click now emit the same error line.

**Do:** the cheap version is to let a step say which kind of failure it is — one optional field on `StepResult`, say `expected?: true`, set on the completion gate and read by `index.ts` to log at `warn` instead of `error`. Setting it on the two demographics branches as well would fix the pre-existing case at the same time, though that touches a live pilot's logging and may belong in its own change. If that's more than this story wants, record in the spec that error volume on all four pilots includes ordinary incomplete answers, so nobody builds an alert on it.

---

## Smaller items

- **The curriculum stage has no end-to-end coverage.** `fall-2026-blue` ([index.ts:63](functions/src/tasks/ai4vs-flvs/index.ts#L63)) is checked only by the handler-identity test. It's structurally spring's tail, so the risk is low — but it's also the stage where `send-email` takes the fallback offering read, and where a missing `email_subject` (R14) produces a notification indistinguishable from a pre-test one. A fourth scenario is a handful of lines: its own `context`, `request: { pilot: "fall-2026-blue", email_subject: … }`, `originClassWord`, and `expect.noAssignment`.

- **Spring's assignment read-back prints a class word that has never existed.** [scenarios.js:50](functions/harness/im-done-local/scenarios.js#L50) builds `fl-spring-2026-origin-shark` from the origin word plus the arm suffix. But spring doesn't append suffixes — it enrols by authored class id, and its real destination is `im-done-shark-202`. Both sides of the comparison are computed the same way, so the check passes while printing a class that doesn't exist. **Do:** assert what the read-back actually reads — an `expect.assignedArm` compared against the stored arm and printed as an arm. That suits the fall scenarios too, where the composed word happens to be real; the word itself is already covered by `enrolledClassId`. A comment saying the composed word is an arm check in disguise is the minimum.

- **The enrollment read-back is opt-in, in exactly the way the last commit stopped tolerating.** Commit `3c986b5` made a success scenario that declares neither `assignedClassWord` nor `noAssignment` fail, on the reasoning that an opt-in branch lets a scenario report PASS while checking nothing. `enrolledClassId` at [run.js:144-149](functions/harness/im-done-local/run.js#L144-L149) has the same shape and no such rule — omit it and `enrollOk` stays `true`. That's the weaker half to leave open, since it's the one assertion that observes the pipeline; the assignment read-back compares two harness values. A Green-stage scenario added later without it would check its completion text and an arm, and nothing about the class the pipeline actually resolved. **Do:** extend the declaration rule to enrolment, so every success scenario states either an `enrolledClassId` or that its stage enrols nothing.

- **A mistyped scenario context falls back silently.** A typo in a `FALL_CONTEXTS` key yields `undefined`, and `{...CONTEXT, ...(scenario.context || {})}` quietly drops back to the shared context — at [seed.js:47](functions/harness/im-done-local/seed.js#L47) and independently at [run.js:95](functions/harness/im-done-local/run.js#L95). The scenario then collides with `happy`'s assignment document, which is R15a's exact hazard. It fails rather than passing, but the failure names nothing useful. Note the stub already guards the neighboring case well: a declared `originClassWord` with no fixture returns a 500 and a message naming the cause, precisely so it doesn't masquerade as a pipeline fault. The context key deserves the same treatment. **Do:** validate every declared scenario once, somewhere both drivers see it — in `scenarios.js` at require time, or a shared helper they both call. A guard inside `seed.js`'s `scenariosNeedingAnswers` loop would cover the two Green scenarios only, missing `fall-orange-control`, which seeds no answers, and any future Blue scenario.

- **`resolveOriginOffering` accepts a `clazz_id` that `lookupClassByWord` would reject.** [portal-reads.ts:144-146](functions/src/tasks/portal-reads.ts#L144-L146) accepts any `clazz_id` that isn't `undefined` or `null`, while the other read in the same file validates ids with `isUsableId` — whose comment says it exists "so the two reads in this file agree on what a valid identifier is." For `null` they do agree; for an empty string or a non-scalar they don't. A `clazz_id` of `""` passes the read, `String("")` is blank, `readStepOutputField` skips it, `send-email` falls back, re-reads the same blank, and posts `class_id: ""`; a non-scalar posts `"[object Object]"`. Both end in a portal rejection rather than a wrong class, and neither is new — but the value now travels as a documented handoff. **Do:** use the existing `isUsableId` in `resolveOriginOffering`. One line, it makes the file's own claim true across the board, and it closes the silent-fallback case that `originClazzId`'s doc comment in [types.ts:21-32](functions/src/tasks/ai4vs-flvs/types.ts#L21-L32) describes.

- **Put the stage in the teacher email.** An enhancement, not a defect. `buildEmailBody` ([send-email.ts:16-36](functions/src/tasks/ai4vs-flvs/send-email.ts#L16-L36)) already prints the student, the offering and one line per step. Adding `Stage: <pilot>` costs nothing, discloses nothing new, and makes a mis-subjected curriculum or post-test notification identify itself. R14's omission is undetectable in code by design; this makes it recoverable after the fact.

- **A stale claim in REPORT-81's spec.** [REPORT-81's spec:187](specs/REPORT-81-pilot-configurable-randomization.md#L187) says the harness's `fall-2026-fulltime` was "the only fall pilot string in the tree," which this story disproves — it appeared in four places, three of them shipped unit tests. This story's spec records the correction; the sentence itself still reads as fact.

PR #409, opened 2026-08-03
fix: stop repeating the step name in evaluate-completion's result message

dougmartin:
`evaluateCompletion` prefixed every one of its five result messages with `evaluate-completion: `. The pipeline already labels each step by name when it renders the results, so the teacher notification email came out reading:

```
- evaluate-completion: evaluate-completion: 4 of 4 questions completed
```

This drops the prefix from all five messages. The step name still appears, once, from the renderer.

Noticed while running the fall-2026 study setup end to end on staging, where the emails are read rather than just asserted on.

No behaviour change beyond the message text: nothing matches on these strings. 338 tests in `src/tasks/ai4vs-flvs` pass.


scytacki (approved):
looks good

PR #412, opened 2026-08-05
fix: shorten CLUE activity names when the unit is a URL [REPORT-98]

dougmartin:
Fixes [REPORT-98](https://concord-consortium.atlassian.net/browse/REPORT-98).

## The problem

`resource_name/1` builds the CLUE activity label from the runnable URL's `unit` and `problem` parameters. MODS deploys its curriculum from a branch, so it passes a whole content.json URL as `unit`, and the label became about 100 characters:

```
CLUE https://models-resources.concord.org/clue-curriculum/branch/modsAuth/mods/content.json: Problem 1.2
```

Eight of the nine resources in the production verification run looked like this. Only the one assignment using a plain `?unit=mods` produced the intended shape.

This reintroduced exactly the redundancy the function's own docstring says it avoids. The comment explains that the runnable URL is deliberately *not* a fallback because "a name repeating it would add a redundant wide column and no information", and a URL-valued `unit` puts the URL back in, in a column that already sits beside `res_N_resource_url`.

## The fix

In CLUE's curriculum layout the directory holding the content file is the unit name, so a URL-valued unit reduces to the same label a plain `?unit=` produces for the same curriculum.

Two details worth reviewing:

**Dotted segments are dropped, not just the trailing one.** Removing only the last segment would leave a versioned directory like `unit.v2/` exposed to producing a filename as the label.

**Anything that yields nothing usable falls back to the unit verbatim**, which is today's behavior. A long label is better than a wrong or empty one, so a bare host or a URL whose every segment is dotted stays as-is rather than degrading.

## Verified against production data

Run through the real `resource_name/1` with the nine runnable URLs from the MODS PD Spring 2026 verification run:

| before | after |
|---|---|
| 104 chars, 8 of 9 resources | **22 chars**, all 9 |

All nine now read `CLUE mods: Problem <n.n>`. The one assignment that already used the plain unit form is unchanged, which is the acceptance criterion (URL-valued and plain forms agree) holding on real data rather than only in a fixture.

## Testing

**515 tests, 0 failures**, up from 510. Five new tests: the URL-valued unit shortens; a URL-valued and a plain unit for the same curriculum produce identical labels; the shortening still applies when there is no `problem`; a URL yielding no usable segment falls back to the unit verbatim; and a relative-path unit is kept verbatim rather than shortened.

That last one covers the form raised in review, `unit=./demo/units/qa/content.json`, used when a unit is served from inside the CLUE repository. It already read correctly, because the shortening only applies to a unit containing `://`, but nothing pinned it. The behavior is load-bearing rather than incidental: shortening it would produce `CLUE qa`, indistinguishable from the unrelated `qa` unit on clue-curriculum, so the test asserts both that the path survives and that it does not collide with a plain `?unit=qa`.

The existing three-case fallback chain (parsed label, unit alone, bare `CLUE`) is unchanged and still covered.

[REPORT-98]: https://concord-consortium.atlassian.net/browse/REPORT-98?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/server/lib/report_server/clue.ex
+++ b/server/lib/report_server/clue.ex
@@ -107,6 +107,13 @@ defmodule ReportServer.Clue do
   ## segments are dropped rather than only the last one, so a versioned directory
   ## cannot leave a filename as the label. Anything that yields nothing usable
   ## falls back to the unit verbatim, which is today's behaviour.
+  ##
+  ## The `://` guard is deliberate, not incidental. A unit served from inside the
+  ## CLUE repository is a relative path, `./demo/units/qa/content.json`, and
+  ## shortening it would label it `qa`, indistinguishable from the unrelated `qa`
+  ## unit on clue-curriculum. Only an absolute URL names the curriculum it came
+  ## from well enough for the directory to stand in for it, so a relative path
+  ## stays verbatim and keeps the two apart.
   defp unit_label(unit) do
     case unit_path_segments(unit) do
       [] -> unit
--- a/server/lib/report_server/clue.ex
+++ b/server/lib/report_server/clue.ex
@@ -83,13 +83,45 @@ defmodule ReportServer.Clue do
   so the runnable URL is deliberately not a fallback: a name repeating it would
   add a redundant wide column and no information. The chain is the parsed label,
   then the unit alone, then a bare "CLUE".
+
+  A unit given as a URL is reduced to its unit name first, so the same guarantee
+  holds when a project deploys its curriculum from a branch.
   """
   def resource_name(runnable_url) do
     query = URI.decode_query(URI.parse(runnable_url).query || "")
     case {blank_to_nil(query["unit"]), blank_to_nil(query["problem"])} do
       {nil, _} -> "CLUE"
-      {unit, nil} -> "CLUE #{unit}"
-      {unit, problem} -> "CLUE #{unit}: Problem #{problem}"
+      {unit, nil} -> "CLUE #{unit_label(unit)}"
+      {unit, problem} -> "CLUE #{unit_label(unit)}: Problem #{problem}"
+    end
+  end
+
+  ## A branch deploy passes the whole content.json URL as the unit, which would
+  ## splice about a hundred characters into a label column and reintroduce the
+  ## exact redundancy the docstring above says this function avoids, since the
+  ## name would then repeat res_N_resource_url.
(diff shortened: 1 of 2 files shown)


dougmartin:
@scytacki This should have probably just been a follow-on commit of https://github.com/concord-consortium/report-service/pull/398 but it ended up its own ticket/PR.

server/lib/report_server/clue.ex
```diff
@@ -83,13 +83,45 @@ defmodule ReportServer.Clue do
...
+  ## falls back to the unit verbatim, which is today's behaviour.
+  defp unit_label(unit) do
+    case unit_path_segments(unit) do
+      [] -> unit
+      segments -> List.last(segments)
+    end
+  end
+
+  defp unit_path_segments(unit) do
+    if String.contains?(unit, "://") do
```
scytacki, line 125: For reference there is another form used for units. When testing against units stored inside of the CLUE code repository we use a url with: `unit=./demo/units/qa/content.json`. It is probably best to just use that full string so it is clear that this is a different qa unit than the one on clue-curriculum. I just wanted to point this out for completeness.

scytacki (approved):
looks good. I just pointed one form of unit value that wasn't explicitly tested.

server/lib/report_server/clue.ex
dougmartin: Good catch, thanks. That form already comes out the way you want, and now there is a test saying so.

The shortening only fires when the unit contains `://`, so `./demo/units/qa/content.json` falls straight through to the verbatim fallback and the label reads `CLUE ./demo/units/qa/content.json: Problem 1.2`. It never reduces to `qa`.

What was missing is that nothing pinned it. The `://` check reads like an incidental "is this a URL?" test, so a later cleanup could reasonably extend the shortening to relative paths and quietly collapse the in-repo `qa` onto the clue-curriculum `qa`, which is exactly the collision you are pointing at. Pushed 9bb4d98: a test asserting the relative path is preserved *and* that it does not equal what a plain `?unit=qa` produces, plus a comment beside `unit_label/1` recording that the guard is deliberate rather than incidental.

515 tests, 0 failures.

